PR #22, opened 2019-01-23
added media tab and page

dzambani:


src/components/Media/index.jsx
```diff
@@ -0,0 +1,25 @@
...
+export default class Media extends React.Component {
+	render() {
+		return (
+			<div className="media-page">
+				<Navbar />
+				<Banner decorative />
+				<div className="content-section center">
+					<h2>Our Media</h2>
+					<p className="subheading">Rad pics.</p>
+						{Config.media.map(media =>
```
alexanderqchen: Looks like the `{Config...}` block was never closed, so I don't think this works

src/components/Navbar/index.jsx
```diff
@@ -55,6 +56,7 @@ export default class Navbar extends React.Component {
 						<div id="hamburger-menu">
 							<ul className="nav-items">
 								<NavLink to="/about"><li>About</li></NavLink>
+								<NavLink to="/media"><li>media</li></NavLink>
```
alexanderqchen: Wait actually good catch here, I forgot about this and I don't think anyone else did this.

alexanderqchen:
And add a description with a screenshot next time!

PR #25, opened 2019-01-23
dzambani/mediapage

dzambani:


src/components/Media/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+	render() {
+		return (
+			<div className="media-page">
+				<Navbar />
+				<Banner decorative />
+				<div className="content-section center">
+					<h2>Our Media</h2>
+					<p className="subheading">Rad pics.</p>
+					{Config.media.map(media =>
+						<p> {media.name} {media.date} {media.tagline} </p>
```
alexanderqchen: no spaces before and after tags `<p>text</>` not `<p> text </p>`

PR #29, opened 2019-01-28
Basic history tile and timeline

ning-hu:
Created and styled the history page with a basic timeline look. 

screen shot 2019-01-27 at 6 19 18 pm
screen shot 2019-01-27 at 6 19 25 pm


src/components/History/Timeline/index.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import HistoryTile from './historytile';
+
+export default class Timeline extends React.Component {
+	render() {
+		return (
+			<div id="timeline">
+				{ this.props.history.map(history => <HistoryTile key={history.title} history={history} /> ) }
```
alexanderqchen: this is a little picky, but remove one space before `)` like this:
`{ this.props.history.map(history => <HistoryTile key={history.title} history={history} />) }`

when i add a linter, things like this should be picked up on automatically

src/components/History/Timeline/style.scss
```diff
@@ -0,0 +1,26 @@
+#timeline {
+	padding-bottom: 20px;
```
alexanderqchen: instead of adding padding to timeline, can you add 40px of margin onto `.history-page .content-section`? That's how it's done for the about and sponsor pages (although I noticed there was some minor errors with our codebase here, so I'll do some refactoring later). Look at `/about/styles` for an example of what i mean.

src/components/History/Timeline/style.scss
```diff
@@ -0,0 +1,26 @@
...
+				text-align: right;
+			}
+		}
+
+		.history-info {
+			text-align: left;
+			border-left: 1px solid black;
+
+			h2 {
+				padding-top: 0px;
```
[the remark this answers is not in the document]
alexanderqchen: Ya that's right. What I mean is that all the other pages have this margin, so the history page's title is like slightly higher up than the titles on all the other pages, which is a little awkward

PR #54, opened 2020-06-28
Refactoring Navbar and Footer out of Individual Pages + Sticky Footer

malsf21:
Hey there!

Had a quick chat with @vohndernet about a small problem with the website a few weeks ago, and this PR should solve that problem.

If you head to [the events page](https://www.uclaacm.com/events) on a tall enough of a screen, the footer won't pad out to the bottom of the screen, and you'll instead be left with an awkward amount of whitespace.

Screen Shot 2020-06-28 at 3 05 13 AM

This PR fixes that problem by making the Footer component sticky via Flexbox.

Screen Shot 2020-06-28 at 3 05 54 AM

Exciting stuff! I should've probably separated this other change, but along the way I also refactored how the Navbar and Footer components work. Currently each of the four pages (`Home`, `Events`, `About`, `Sponsors`) each individually imports `Navbar` and `Footer` and renders them inside their own page, but uses them identically. This PR moves it up to `main.jsx`, which should technically lighten each page load (albeit minimally). 

I had to do this to make the sticky footer work properly (as the footer needs to be a sibling element of the "main" app content). 

Let me know if there are any problems with this refactor + sticky footer!

### DIFF
--- a/src/components/About/index.jsx
+++ b/src/components/About/index.jsx
@@ -2,8 +2,6 @@ import React from 'react';
 import Config from 'config';
 
 import Navbar from 'components/Navbar';
-import Button from 'components/Button';
-import Footer from 'components/Footer';
 import Banner from 'components/Banner';
 import Officers from 'components/About/Officers';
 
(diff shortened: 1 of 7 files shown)


vohndernet:
Super clean, @malsf21! Good to merge :)

PR #56, opened 2020-07-02
Simplifying & Standardizing the Dev + Deploy Pipeline, Major Site Refactor

### LINKED ISSUE #14: Implement cache busting with webpack (requested by Christina Indudhara)
## Motivation
We want to make sure users have the newest version of the website especially when important content is changed.

## Resources
https://stackoverflow.com/questions/39238163/how-can-i-make-webpack-use-a-cache-busting-suffix

## To Do
- [ ] Make sure that when the website is updated, the browser does not simply fetch the files from cache
- [ ] Make sure that if the website is not updated, the browser does get it from cache

malsf21:
Okay, this is the big PR! As I've mentioned earlier in discussions with @vohndernet and other ACM peeps, there's a bit of overkill in the current CD process: things like Docker, heroku, and homebrewed webpack scripts are great, but they add a lot of cognitive load to a dev team that's already pretty strapped for resources. 

This PR commits the *cardinal sin* of using [Create React App](https://github.com/facebook/create-react-app) instead, to manage a large portion of the development and deployment pipeline. This is the largest breaking change, and requires some retooling of our current dev resources. However, after this PR is merged, we can move away from using AWS/heroku to static site deployers like Netlify or GitHub Pages.

However, we made sure that this wasn't a regression. In moving to CRA, we may lose some benefits of having more control over our system. We can resolve that in a few ways:

* Do you still want to customize webpack? There are webpack entrypoints for CRA!
* Do you want to fine-tune our build process? Instead of brewing our own solution, we should use a CI/CD service like GitHub Actions, Travis CI, AppVeyor, etc. We can still specify software images, node versions, and all that goodness.
* Do you still want to gzip your files and optimize for a static site deployment context? CRA's webpack config already does some of this, but we can also configure these types of options in the Netlify console (should we proceed with them)
* Do you still want to customize babel, ESlint, etc? That still works with CRA!
* CRA does cache-busting for us with hashed JS! This can closes #14 without developer intervention.

In addition, this PR also cleans up many issues in the resulting JSX (including lack of `alt`, misuse of `target="_blank"`, overloading `style`, overloading `Number`, and others that are detailed in the changelog). Also, since we're in the process of refactoring anyways, we move all trivial components to functional components, and do some renaming to be more consistent with typical React projects. 

In theory, this can actually be merged right now (assuming the deployment is reconfigured), with no repercussions - we have introduced no code regressions, and the website still functions as normal (and in some instances, better) than what's on live/`master`.

Stuff I still want to do before this is *actually* mergeable:

- [x] set up CI/CD pipeline with smoke test for compilation
- [x] fix footer spacing (no blank headers!)
- [x] correctly audit the use of images with `PUBLIC_URL`
- [x] document changes & new workflow

Short term things that we should do:

* Cleanup complex JS classes
* SCSS passover and namespace overview (there are some strange style choices, such as `Button`)
* rename `Config` to `config` (there actually is a general collision on `config`, which is bad)
* Properly use Favicons (there is no ICO currently, + generating array of different favicons)
* refactor the carousel (it should be using the `img` tag!)
* refactor the `<Button>` component (specifically, we shouldn't be wrapping a button in an `a` tag like in the footer)

Also, this lets us do a few really really cool things:

* now, we can write some tests for parts of our website, especially things like helper functions and layout generators
* we should start adding JSDoc for our components and functions!
* typescript?! type annotations? that would be great :smile:
* use Hooks properly! since now more components are functional!
* probably do some serious refactoring and redesign, now that there's much less clutter going on
* investigate service workers!

Anyways, this is a lot. Let me know what you think! In the meantime, you can give this a spin with

``` 
$ npm install
$ npm start
```

### moving to create-react-app

This is one of those changes that's really "we just did a lot of work, there's no diff-ing". Here's my attempt:

* installs `react-scripts` (CRA), moving previously-explicit packages to CRA's control, including:
  * `webpack`, related packages (`css-loader`, `sass-loader`) and config files
  * `babel`, related packages (`babel-*`), and config files
  * `eslint` and config files
  * moves HMR to CRA instead of `webpack-dev-server`
* removes now-unused artefacts from current build pipeline
  * removes docker artefacts
  * removes Nginx artefacts
  * removes `Makefile`
* moves files from `static` to CRA's `public`
* updates `index.html` with new relevant build information
* removes unused deps
  * `npm` (why was this here)
  * `tar`
  * `lodash` and `lodash-es`
* updates `homepage` property to use the right URL
* adds GH Action to build on commit
* documents changes and updates README to have more contribution information

### refactoring to meet deploy standards

* when possible, refactors stateless components into functional components
  * this is every component except `Banner`, `Carousel`, `Navbar`
* refactors the `style` prop on `Button` to `styleClass` (as `style` is a reserved prop) -> eventually should fix root problem, in SCSS definition of class or properties of component
* renames all `*.jsx` files to `*.js`
* renames all files of form `Component/index.js` to `Component/Component.js`
* fixes spelling error in "carousel/Carousel" (prev "carousal/Carousal")
* fixes general accessibility issues
  * many, many missing `alt` tags
  * adds `rel="noopener noreferrer"` to all uses of `target=_blank`
* removes hacky `Number.map` overload, replaces with a helper function - might be an easier way to do this
* uses the public URL for all iamges
* writes a bandaid fix for spacing out the mailing list button in the footer without using an empty header tag. should be re-adjusted as `Button` is refactored.


malsf21:
@vohndernet completed all the must-do tasks to merge. Let me know what you think!

(this is blocked by activating our static site deployment and setting up our DNS records)

PR #59, opened 2020-09-14
Nvan/updates

vohndernet:
Miscellaneous Updates (September 2020)

Changes are categorized under the pages they belong to. Home page has the most obvious-looking changes, a la "In the News" section, but I kept it closely to how it looked before! If we prefer a greater change, though, then I can also make it look much less closely.

**HOME**
- [x] rename RECENT ACHIEVEMENTS to IN THE NEWS
- [x] update IN THE NEWS formatting (rounded image borders, clickable articles, new articles)
- [x] Increase font size of Sign-Up region (Join our Mailing list), though looking to still work on making mailing-list button more prominent...

**FOOTER**
- [x] add youtube button
- [x] resize buttons for uniformity 
- [x] make email clickable

**ABOUT PAGE**
- [x] add icons row to section "How do I get involved?"

**EVENTS**
- [x] remove spring-quarter COVID tagline

**SPONSORS**
- [x] make icons clickable and lead to company website
- [x] updated new logos for certain companies


chern:
Looking good so far! A few things:

- Another thing I thought of — could we switch the body font from Chivo to [Open Sans](https://fonts.google.com/specimen/Open+Sans)? I think it'll look better
- Make the drop shadows on the "in the news" cards lighter? Or maybe eliminate them altogether and instead put a a thin outline around them instead? The heavy shadows look weird and like they're from the early 2000s haha
- Change from all uppercase to title case for the "in the news" card titles? All caps is a little harder to read
- Let's add a button to our YouTube channel on the events page?

That's all I can think of for now — fantastic work thus far!

austinkeil (changes requested):
Great job, the site's looking beautiful; thanks Nhi!

Agree with William's suggestions, some other ideas:
- The blue underline on the "reach us at acm@ucla.edu" in the footer doesn't have enough contrast with the background, maybe we can use the ACM Tint [RGB: (172, 186, 250), hex: ACBAFA] from the design style guidelines rather than ACM Cobalt?
- In line with what William said about the drop shadows on the In the News section, we could use the same borders that we use for the officer images on the about page (border: 2px solid #EAEAEA) with no drop shadow and I think that might look cleaner and more consistent across the site

Great stuff!

chern (changes requested):
A few more things I thought of:

- For the "In the News" cards, let's make the articles open in a new tab when clicked
- For the committees on the front page, we should either arrange them by color (like in a rainbow) or alphabetically; the current ordering seems arbitrary
- Let's take the row of social media buttons you added on the "About" page and also put them on the front page, right above the committees, to make the social media links more prominent
Screen Shot 2020-09-14 at 12 12 33 PM

vohndernet:
Thanks so much for your feedback! I've pushed the following changes below to address them:

GENERAL
- [x] switch the body font from Chivo to Open Sans

HOME (Committees)
- [x] arrange the committees by color (like in a rainbow)
- [x] Added social media buttons row to Home page, right above the committees 
- [x] was not discussed here on the pr, but I also updated the discord link, to reflect a recent PR merged by Evan [here](https://github.com/uclaacm/website/pull/60)

HOME (News)
- [x] put a a thin outline around the images, similar to the borders used for the officer images on the about page
- [x] Change titles from all uppercase to title case
- [x] articles open in a new tab when clicked

EVENTS
- [x] add a button to our YouTube channel on the events page

EMAIL (Footer)
- [x]  blue underline under email uses ACM Tint as opposed to the current dark blue

chern (changes requested):
Great work, Nhi — it's looking really good! One thing I updated is the Google Fonts stylesheet for the font update, too.

Two more things:
- Could we use ACM blue (3B59ED) for the social media button colors? This way it'll stand out on the page more and help indicate what's clickable/tappable
- Could we add the 2019 ICPC qualifies for North American championship and 2017 + 2018 chapter excellence awards to the "In the News" section? Feel free to DM me if you need anything from my end for that

Overall, looking really good! Just a couple minor things here and there, and then we should be good to go!

vohndernet:
Thank you for updating the stylesheet, Will! I've implemented the two updates you suggested for the SocialMedia icons and the News section, as well as some refactoring of their respective coding while I was at it, just to make it cleaner. You're right about the ACM blue change–– it really makes the icons more prominent.

Let me know if there's anything else I should work on!

chern (approved):
Looks awesome, Nhi! I tweaked a few news headlines and adjusted a few image sizes to conform to a 16:9 aspect ratio so that the images don't look stretched. Everything's looking great to me!

The only last thing I can think of is if there are any tweaks we can make to display the mailing list signups more prominently — but we don't have to do it in this update if we're dry of ideas at the moment. These changes look good to me, and I think it's ready to merge unless @austinkeil has any other suggestions!

austinkeil:
Looking great, Nhi! Thanks for all the hard work!

PR #62, opened 2020-09-25
Wchern/carousel-updates-2020

chern:
Carousel updates:

- Add 24 newer, more recent photos (from the 2018-19, 2019-20 school years) to the carousel (for a total of 40 photos)
- Speed up the carousel scroll animation
- Adjust image appearance in carousel


vohndernet (approved):
The new pictures look great, as do the other changes! At first I wasn't sure about the hover change from increasing opacity to decreasing opacity, but now I realize it's because of how the mailing list banner above frames the carousel too much that it detracts from the effect... I'll keep brainstorming on that front. Thanks for your hard work, Will! LGTM

chern:
Thanks, sounds good! I mainly wanted to adjust the image appearance because I wanted the photos to pop more (kinda didn't like dimmed look of the photos previously), but definitely feel free to improve as you see a better way

PR #63, opened 2020-09-26
Fall gm 2020 virtual program (part 1)

evazhog:
image

This is the first part of the Fall GM 2020 virtual program. The 2nd part will follow in a later PR.

The idea is that we are adding a new (temporary) page to the ACM website under routing `/fall-gm`. 
- Up until 6:30 PM PDT on Monday October 5th there will be this countdown displayed. 
- After 6:30 and until we kill the page (upon conclusion of GM) the page will display a "virtual program" for GM (what's happening at what time) as well as links to join the Zoom call and Discord social. 

This "virtual program" will follow in another PR shortly

Ideally Operations would like this deployed to production by Saturday night, but we can work with a timeline where this is deployed midday Sunday. We were planning to link to this page in the ACM newsletter (sent out Sunday evening) as well as the ACM facebook page (which we will be pushing over the weekend)


src/App.scss
```diff
@@ -59,6 +59,7 @@ body {
 }
 
 .main-container {
+	margin-top: 70px; //To account for height of Navbar, which uses fixed positioning
```
evazhog, line 62: This is a small refactor I did.

Previously the decorative Banner was responsible for having extra margins to account for the thickness of the nav-bar (which is stuck using fixed positioning).

This refactor moves the responsibility to the main-container (which is a high level parent of all pages). This makes adding new elements more intuitive, as all content should be wrapped within the main-container. Previously adding new elements would appear at the top of the page, but the content would be hidden behind the fixed positioned nav-bar

src/components/Banner/style.scss
```diff
@@ -75,7 +74,7 @@
 
 	.title {
 		position: absolute;
-		top: 50px;
+		top: 30px;
```
evazhog, line 77: Same refactor as above

src/components/Events/Fall_GM/Fall_GM.js
```diff
@@ -0,0 +1,64 @@
...
+export default class Fall_GM extends React.Component {
+	constructor(props) {
+		super(props);
+	}
+
+	render() {
+		const FALL_GM_2020_START_TIME = new Date('2020-10-05T18:30:00-0700'); //October 5th 2020 at 6:30 PDT 
+	
+		const renderHumanReadableTime = ({ days, hours, minutes, seconds, completed }) => {
+			if (completed) return <h1 id="time-to-gm-headline-now">Fall General Meeting 2020 is happening right now!</h1>;
```
evazhog, line 15: This is where the program will go in the next PR

src/components/Events/Fall_GM/Fall_GM.js
```diff
@@ -0,0 +1,64 @@
...
+							<div className="countdown-labels">{minuteString}</div>
+						</div>
+						<div className="countdown-cards">
+							<div className="square-background"/>
+							<div className="countdown-numbers">{seconds}</div>
+							<div className="countdown-labels">{secondString}</div>
+						</div>
+					</div>
+					<p id="cant-wait">Can't wait to join in on the fun? Jump in right now with our <a href="https://linktr.ee/acmucla">Linktree</a> or hop into the official <a href="https://bit.ly/ACMdiscord">Discord</a> and say hello!</p>
+				</div>
```
evazhog, line 53: This probably could have been componentized but I decided against it for readability.

As this function becomes more complex (due to the addition of the virtual program) I'll move both of them into separate components.

evazhog (commented):
added comments for clarification

austinkeil (approved):
Great stuff. Will likely need to reduce the length of the entry in the navbar from 'fall general meeting' to 'fall gm' as we're also adding a committees page and all of the entries won't fit. Don't worry about updating anything, I'll make the change when we add the committees page if it's needed. Thanks for the good work!

PR #65, opened 2020-09-27
Fall gm 2020 virtual program fast follow #2 (better mobile support)

evazhog:
Hakan and Gloria pointed out that the countdown timer on the new `/fall-gm` page does not offer enough support for mobile users.

This small patch attempts to mitigate that problem (and others)
- include a link to `/fall-gm` in the hamburger nav bar, which see high use from mobile users
- allow countdown to gm component to wrap depending on screen size
image
previously the countdown would overflow the viewport:
image
- resize fall gm graphic to fill more dead space on mobile devices
- change the `max-width` that determines when the nav-bar changes to hamburger menu (avoids the following UI overflow, thanks to @chern and @malsf21 for this one)
image


src/components/Navbar/style.scss
```diff
@@ -198,7 +198,7 @@
 		color: $acm-cobalt !important;
 	}
 
-	@media (max-width: 700px) {
+	@media (max-width: 875px) {
```
evazhog, line 201: Note @austinkeil:

This is due to the length of the text `fall general meeting` in the nav-bar. When we shorten it to `fall gm` and add the about committees section, this responsive max-width will likely need to change again.

chern (commented):
Minor issue with the footer: "Find us on social media" and "Reach us at" not correctly centered (not sure when this started, but a screenshot from right before all the fall GM program commits show that it was still correctly centered then)

Screen Shot 2020-09-27 at 10 48 44 AM

chern (approved):

PR #69, opened 2020-10-01
Affiliates updates: Fall 2020

chern:
Update sponsors page with latest list of departmental affiliates


vohndernet (changes requested):
Since Symantec went under quite a bit restructuring with the Broadcom purchase the past year, I don't believe they go by that name anymore, unless they wish to do so for this particular sponsorship? Can we update the logo/link to either Broadcom or NortonLifeLock? Otherwise, everything else looks good.

chern:
Good catch — we're still trying to clarify that with the department, but for now External has advised that we keep it as Symantec

vohndernet (approved):
That makes sense. LGTM

PR #73, opened 2020-10-04
Committees

austinkeil:
# Description

Add committees page to ACM website. 

*The sections with images look much better than those without; harassing committees for more images is encouraged.

ezgif com-video-to-gif

## Summary
- Social icons change color depending on the committee section using CSS masks ([supported for 97% of users](caniuse.com/css-masks))
- 688px used for breakpoint for tablets, 992px for desktop devices
- min-width used for media queries following the mobile-first convention
- Updated social icons to square versions (for companies where this style is supported) to meet ACM Design's recommendations
- Added new photos for each committee
- Added several new components for use in committees page (CommitteeBanner is the banner for the whole page; there is a Header, Intro, and Body for each committee section, and this relationship is shown in the file hierarchy)
- Navbar collapses at different breakpoint due to change in content's width
- Updated config/committees to support a lot more data used in committee page

## Testing:
- Changes generate no new warnings
- Looks good on mobile, tablet, and desktop screen resolutions (navbar hides correctly, views change appropriately)

## Notes:
- Might want to split config/committees into its own config (separate from the one used by the homepage) so we're not passing so much irrelevant info? 
- We should develop some official style guidelines + a linter for our projects.


chern (changes requested):
Looking great, Austin! Two things:

- Every title under an image card looks slightly off-balance because it's not quite lined up with the left margin (e.g. "CodeSprintLA" and "Cyber x Hack x …" are slightly off from the image's left margin)
Screen Shot 2020-10-03 at 11 05 07 PM


- Could you update the "ACM [committee]" wordmarks tomorrow after the rebranding launches?
Screen Shot 2020-10-03 at 11 02 53 PM
Screen Shot 2020-10-03 at 9 54 11 PM

chern (changes requested):
Also, I don't think we need "Mission: …"; we can just put the committee's description directly in there

Screen Shot 2020-10-03 at 11 08 49 PM

vohndernet (changes requested):
Beautiful work, Austin! I agree with Will's suggestions, and have a few of my own (some style, some content)

- To add to Will’s comments about the rebranding, we can also replace the first banner image as well with the rebranded version. (for reference, it’s in the about page, now that the rebranding has been launched!!)
- Could we have the social media icon for a committee’s website be their diamond instead? I feel this would look more appealing and clickable than the current www icon in use.
- Since the different image sizes make it hard to keep the events section neat, it would help to have a fixed height for the images, even if they can't have the same ratios. I also think with certain images, even resizing won’t be an issue. For example, cyber’s, w’s, and hack’s image size differences are slight enough that minor adjustments (or even cropping) would preserve the image well even if you forced them to have the same ratios there.  
- There’s an instance of “Cyber x Hack x ICPC presents Bruin Quest” under all three participating commitees, but ICPC’s event does not have the same picture! (it’s using the default icpc cover). We should update that to also have the same photo.
- For CLxACM Hack presents Bloom, I believe they have that facebook cover you can use for the event photo. 

bloom_hack

- This is more a content concern than a style one, but for the event descriptions, it’d be neater if we shorten the links that committees provided there.  I don't think the committees would mind if you make this kind of adjustment. Writing out the full url can be a bit distracting, so having it be a text-hyperlink would be more streamline. e.g. "You can see a collection of many past projects at [uclaacmstudio.itch.io](uclaacmstudio.itch.io)." —> "You can see a collection of many past projects [here](uclaacmstudio.itch.io)."

src/components/About/style.scss
```diff
@@ -80,10 +80,6 @@
 	}
 }
 
-
-@media (max-width: 750px) {
-}
```
evazhog, line 85: Assuming this is just removal of a useless media query?

src/components/Committees/CommitteesBanner/CommitteesBanner.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default class CommitteesBanner extends React.Component {
+	render() {
+		return (
+			<div className="committee-banner-container">
+				<div className="banner-main-content">
+					<div className="banner-left-col">
+						<img src={process.env.PUBLIC_URL + "/images/acm_committees.svg"} alt="the logos of all ACM committees" />
+					</div>
+					<div className="banner-right-col">
```
evazhog, line 11: I'm a little confused with this className, doesn't seem to be used in the associated `.scss` file

src/components/Committees/CommitteesBanner/style.scss
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+    @media (min-width: 688px) {
+		min-height: 600px;
+		height: 50vh;
+
+		.banner-main-content {
+				position: relative;
+				top: 50%;
+				transform: perspective(1px) translateY(-50%);
```
evazhog, line 19: I'm unclear as to what this CSS does. I tried removing the relative position and the element looked like it rendered in the same way

src/components/SocialMedia/style.scss
```diff
@@ -3,8 +3,8 @@
     width: 30px;
     border-radius: 2px;
     margin: 0px 3px;
-     display: inline-block;
-    //color: white;
+    display: inline-block;
+    // color: white;
```
evazhog, line 7: Pls remove comment before merging

evazhog (changes requested):
Great work Austin!

Left some comments on specific files. My main takeaways are:
- Most new components don't require state and can thus be made into functional components
- Issues with differing aspect ratios in "Events and Projects" images. We can potentially solve this using CSS masks like you mentioned image
- Potentially confusing UX: We use the same "image with border and text below" on our homepage when linking to relevant ACM new. Those components are wrapped by `<a>` tags that users can click and navigate to external pages. On `/committees`, however, the components are not interactive. This is very minor but users may attempt to interact with the components on our committees page. One solution is to make the components on the homepage more interactive (dim if cursor is hovered) which makes them more distinct.

|/|/committees|
|---|---|
|image | image|

- Make naming as descriptive as possible. As our codebase grows, making component names as descriptive as possible makes navigating easier.

Love to see it. Hope to review this again one the rebranding changes have been absorbed in

src/components/Committees/CommitteesBanner/CommitteesBanner.js
austinkeil: Ah, originally had gone with a two-column layout for larger screen sizes but decided not to go that route; good catch

src/components/Committees/CommitteesBanner/style.scss
austinkeil: This just centers the image vertically in the parent container. For very large screens this has more of a noticeable effect; I've changed the way this is done in the new commit

austinkeil (commented):
All requested changes have been made barring shortening the URL links which @vohndernet is working on now! Thank you for your reviews, all great feedback.

austinkeil (commented):
Changes made: thanks, Evan!

evazhog (approved):
Beautiful work, lgtm.

vohndernet (approved):
Committed the shortened links. Minor rewording to accommodate it, but we're all set to go :)

austinkeil (commented):
Great stuff @vohndernet!

chern (approved):
Great stuff y'all, let's merge it in time for Fall GM tabling tonight… one more tweak that I'd like to see though (just spoke with Austin about it)

PR #71, opened 2020-10-02
Nhvan/branding update short

vohndernet:
This PR is for the short-term rebranding and after review will be launched on Sunday.

The major highlights, include
- [x] ACM logo and general colors
- [x] committee gradient logos
- [x] new committee colors, as seen in Banner 
- [x] updated About page with image of new committee logos

In all regards, this is a straightforward replacement PR, but any feedback on other stylistic changes would be greatly appreciated!

PR #74, opened 2020-10-04
Rebranding fast follow

evazhog:
A minor fix that may have fallen through the cracks during rebranding on the `/about` page:

- updated `.square-splash`, `.square-ornament`, `.square-small`, and `.square-tiny` with the new border radii convention

|  Previous  | This PR  |
|---|---|
|  image |  image |

Unresolved but noted issue:
The image next to the how to get involved section is a static png, so will need to be updated in a photo editor and cannot be done through CSS
image

### DIFF
--- a/src/components/About/style.scss
+++ b/src/components/About/style.scss
@@ -7,7 +7,7 @@
 .square-ornament {
 	height: 200px;
 	width: 200px;
-	border-radius: 20px;
+	border-radius: 40px;
 	background: $acm-tint;
 	transform: rotate(45deg);
 }
@@ -39,7 +39,7 @@
 	width: 250px;
 	left: 30px;
 	top: -65px;
-	border-radius: 25px;
+	border-radius: 50px;
 	background-image: url(/images/about1.png);
 	background-size: cover;
 	background-color: $acm-cobalt;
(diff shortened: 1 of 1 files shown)


src/components/About/style.scss
```diff
@@ -7,7 +7,7 @@
 .square-ornament {
 	height: 200px;
 	width: 200px;
-	border-radius: 20px;
+	border-radius: 40px;
```
evazhog, line 10: For all of these updates I've assumed the ratio of 20% border-radius w/ respect to width (40/200)

I'm not sure this is correct, however, b/c I've seen other ratios used in #71 such as 15/50 and 40/150. Defer to @chern @vohndernet @austinkeil as they know the most about the rebranding style guide.

We should const-ify this somehow as the `border-radius` css property takes percentages as a value. Like so:
```
App.scss:
$diamond-border-radius-percentage: X%;
---
Other.scss
.square {
    width: Ypx;
    border-radius: $$diamond-border-radius-percentage;
}
```

austinkeil: This is a really good point, let's get this PR through and @vohndernet and I will work on putting a standard in place and raise the change in another PR.

austinkeil (approved):
Good work and great observations: adding the changes to our next sprint.

PR #75, opened 2020-10-04
Tweak corner radius

chern:
Tweak the height-to-radius ratio to be 0.28, based on guidance from Design


vohndernet (approved):
Design knows best :) LGTM

PR #85, opened 2020-10-09
Committees Page Updates

austinkeil:
- compressed all images on committees page to improve loading speed
- update 'learn more' button on home page
- update youtube link to new custom url
- added new photos from various committee websites and Facebook groups


chern (changes requested):
Thanks for being on it, Austin! One minor thing, this is the old motif from the old branding and not really specific to the workshop, so we should still continue to use the committee's placeholder here instead:

Screen Shot 2020-10-09 at 5 19 43 PM

chern (approved):
Looks good!

PR #96, opened 2021-02-09
Updated committee info for Design and Cyber

tnawathe21:


vohndernet (changes requested):
Great work, Tanaya! Just a few changes for you before we get it through:
Screen Shot 2021-02-08 at 10 29 35 PM

- There’s a third button here, but Design only has the website and facebook! The confusion is probably because their styleguide is already their website, so you don’t need to have this extra button to separate them. (remove this extra button please)
- regarding the first photo for Design
    - it doesn't show up! hint: the png is labelled rebranding_udpate.png while in the json config, it’s rebranding_update.png
    - change the alt tag to reflect the new photo! It’s no longer a ‘Team photo’
- Currently our Learn More buttons are intended for social media platforms, and not emails, so it won’t work if you use this kind of button for Cyber’s email (remove please)

You're doing great so far :) Keep it up!

vohndernet (approved):
Great! Feel free to merge into the master branch :)

PR #99, opened 2021-03-03
Tech Gala Projects W21

julia-yin:
Created the Tech Gala Projects page for Winter 2021. Added Tech Gala banners to the Home and Events pages.


src/config/tgprojects.js
```diff
@@ -0,0 +1,89 @@
+export default [
+    {
+		title: 'Covalent',
+        names: 'Arek Der-Sarkissian, Daniel Dai, Muthu Palaniappan, Will O\'Brien, Hritik Arya, Catherine Hu, Daniel Lee, Michael Shi',
+		summary: 'We\'re helping teams build better bonds with virtual icebreakers.',
+        category: 'Games & Graphics',
+		image: '/images/techgala/projects/covalent.png',
+        winner: 'Winner!'
```
evanzhong, line 8: I realize that our spreadsheet didn't specify what type of award these teams won, that's on me.

Mind updating this to 
```suggestion
        winner: 'Best Large Team'
```

src/config/tgprojects.js
```diff
@@ -0,0 +1,89 @@
...
+	},
+    {
+		title: 'Hotseat',
+        names: 'Timothy Gu, Nathan Smith',
+		summary: 'Hotseat is an all-in-one course discovery website for UCLA students. With Hotseat, you can track courses and receive real-time notifications when seats open up, see grade distributions of previous sections, and see historical enrollment data in order to know which sections to use your first pass units on, among many other features.',
+        category: 'Education',
+		image: '/images/techgala/projects/hotseat.png',
+        winner: 'Winner!'
```
evanzhong, line 16:
```suggestion
        winner: 'Best Small Team'
```

src/config/tgprojects.js
```diff
@@ -0,0 +1,89 @@
...
+	},
+    {
+		title: 'Probabilistic Graphic Generator 2020',
+        names: 'Ellie Cheng, Eric Wang, Bill Liu, Lucas Xia',
+		summary: 'We created a graphic generator using a probabilistic factor graph model and a modification of Metropolis-Hasting sampling to automate the process of creating simple yet aesthetically pleasing graphics that can produce graphics for desktop backgrounds, T-shirts, mugs, posters etc. where details are not as integral, and randomness is welcome.',
+        category: 'Design',
+		image: '/images/techgala/projects/probabilistic.png',
+        winner: 'Winner!'
```
evanzhong, line 24:
```suggestion
        winner: 'Best in Show'
```

src/components/TechGala/TechGalaBanner/TechGalaBanner.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+function TechGalaBanner(props) {
+	return (
+        <div>
+            <img src={process.env.PUBLIC_URL + "/images/techgala/techgala-banner-light.jpg"} alt="2021 tech gala banner" />
```
evanzhong, line 6: Would it be possible to use the dark-mode banner here? It's the final marketing we went with, though I do realize that it has time and date info that might not be entirely relevant since the event has already passed.
image


```suggestion
            <img src={process.env.PUBLIC_URL + "/images/techgala/techgala-banner-dark.jpg"} alt="2021 tech gala banner" />
```

evanzhong (commented):
Looking good! 

I reviewed mainly on design and UI choices, and only gave a cursory glance at the code (but that looks well done as well). I left some comments on specific sections of the diff.

Some takeaways I have:
 - The layout is narrow, while mobile friendly, it feels a little cramped when viewed on a widescreen. Can we try a 3-column layout? I messed around with the inspector to get a rough mock (of course we could move thing around like the images, text, names etc to get everything to fit more smoothly) image
 - I like where this banner is going, we just need to figure out where in the layout it belongs. The way we currently position `<Banner>` component is very hacky, with negative margins etc. so finding a good solution here will be tough. image image
 - We don't link to this page in the nav-bar so it's difficult to find unless you know where to look. I realize this might be a conscious choice, however, as our nav bar is getting a little full.

vohndernet (changes requested):
Great work so far, Julia! Got a couple of suggestions for you before we can get this up:

[**Issues to fix**]
- The images on the committees page have grey borders when they shouldn’t: 
Screen Shot 2021-03-03 at 9 20 25 PM

This is probably an issue with either your css or how you define the image items on your Tech Gala page, so it bled out to the images on the committees page. This will need to be fixed before the page can be deployed.

[**Suggestions**]
- ~~can we change the Winner! labels to the individual places instead?~~ EDIT: Ignore this and go with Evan's suggestions on them
- It’d be better to make the tags a different color than black, such as one of our pre-defined ACM blues, which you can find defined in `App.scss`.
- For the few projects that did provide us links, we should add them as well if possible.

**[Design-wise]**
- considering the time-sensitivity, this current design is overall good to be deployed. Some improvements we can consider design-wise (now or eventually) would be having the 1st/2nd/3rd place projects out in a row at once, or some greater distinction to single them out as winners (maybe a more graphical pedestal?) 
- Also, if we get a larger amount of projects than currently, scrolling through would take a long time. Some options would be changing it to a more grid layout, adding pagination within the page, or something else more compact. 
Again with the time crunch, it's fine to save most of this design section for a later date. [EDIT: submitted this review before i realized you reviewed already my bad evan LOL]


And adding another comment here,  just so it's noted down––regarding how to access it, the tech gala page is currently accessible through a banner on the home page or the events page. This is great! This does though ask the question of how to handle future events pages, such continually adding more items in the nav bar (such as winter gm) or adding more banners (tech gala), neither of which is sustainable. Eventually we should either modify the events page to be more of a landing page for all individual event pages, or by making its nav item as a dropdown for all events. This can be another future ticket for the website.

Overall, loving everything you're doing, Julia <3 Keep up the great work!

mattxwang:
Hey, just wanted to check in - @julia-yin  or @vohndernet, what's blocking this PR right now? Anything on our end we can do to help?

mattxwang:
resolved the conflict (caused by me this time around, mb!)

src/components/TechGala/Projects/ProjectCard/ProjectCard.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import Tag from './Tag/Tag';
+
+function ProjectCard(props) {
+    return (
```
evanzhong, line 5: This is perhaps personal preference, but I'm a big fan of "object destructuring". I think it makes code more readable and makes things DRYer (DRY is an acronym for "Don't Repeat Yourself").

For example you can do

```suggestion
const {image, title, winner, summary, names, url, category} = props.project;
    return (
```
to create a bunch of local variables holding the information you want, then you can simply reference these variables instead of the `props.project.` object in your returned HTML.

evanzhong (approved):
Hey Julia!

Everything looks great! I've left some minor comments in individual files. None of them should block you from merging this but here's the overview:

- We can make your code more readable using object destructuring
- Considering the tradeoff between performance and reducing file pollution by using inline-CSS 
- We usually make sure all files have an empty line at the end (as git convention). There's no need to manually do this now, as we can add an eslint rule and run it on these files later.

Great work!

mattxwang:
In general I'm good to merge, but looks like we'll have to resolve a few merge conflicts before moving ahead. We can work on this in today's meeting, or later!

PR #101, opened 2021-04-01
Updating "Officers" With New Leadership

### LINKED ISSUE #100: Update officers section w/ new leadership
Now that almost all the cprez + PVP are confirmed, we can update the respective datafiles! This is pretty light, so one of us can handle it.

mattxwang:
Replaces new PVP + 6 out of 8 committee presidents (not including vice presidents); updates to Cyber and ICPC happen mid spring-quarter.

Also introduces a "breaking change", removing emails from the JSON. However, note that we didn't actually use this field anywhere!

Draft as we review photos; for reference, sourced them from:

* ACM Directory
* LinkedIn
* Facebook

Closes #100.

P.S. we may want to update the design / layout of this section, but that's for another PR 😉


mattxwang:
Just got the go-ahead to merge this - if anybody on dev can review, I think we're good to go!

vohndernet (approved):
Looks good to me! Making a note that we should fix the alignment between the picture and the name text (long names make it offcentered because it goes to the next line), but as you said, that can be done on a future ticket. LGTM!

PR #108, opened 2021-04-09
Created Nav Bar for Committees Page - addresses issue #104

### LINKED ISSUE #104: Add Navigation Bar on the Committees Page
The commitees page is both great and rather long to scroll through if you're not looking through every single committee (what if you just want to see facts about ACM-W?). We should implement some navigation, so that others can look the committee they want more easily. 

What I'm thinking of is a navigation bar (placement on the left side), where we have the committee diamonds you can click on, which will take you to the committee section. 

Some Examples
- TeachLA's [CipherSalad](https://ciphersalad.uclaacm.com/) has the functionality I had in mind, with its keys that navigate to each step. Instead of keys, though, we should use the different committee diamonds instead. 
- SoftServe's [website](https://www.softserveinc.com/en-us)'s navbar has more special effects we could consider adding, like the words that appear on-hover. 

Feel free to ask if you have any questions, or have comments!

tnawathe21:
closes #104


src/components/Committees/CommitteesSidebar/style.scss
```diff
@@ -0,0 +1,42 @@
...
+    z-index: 1000;
+    padding: 1rem;
+    transform: translateY(-50%);
+    display: flex;
+    flex-flow: column nowrap;
+    //justify-content: space-around;
+
+	.committee {
+		width: 200px;
+		height: 63px;
```
mattxwang, line 16: Giving a fixed width to each of the images is a bit tricky; you might notice that each diamond isn't scaled the same (ex the "smaller" names for W and AI are much larger than longer ones like Teach LA or Studio). 

Instead, I would recommend:

1. left-align all the committee images, so the diamonds are stacked on top of each other
2. fix the height of both the image and its link container
3. fix the width of the container for each committee, but not the image itself - this should let the image fill as expected 

A side effect of this approach is that it may help you reduce some classes (ex using `display:block` explicitly, not too many `margin` problems). Let me know if this makes sense, this is a tricky review comment and I can walk it through with you 1-on-1!

src/components/Committees/CommitteesSidebar/style.scss
```diff
@@ -0,0 +1,42 @@
...
+		display: inline-block;
+		color: $font-primary-color;
+		text-align: left;
```
mattxwang, line 19: Not sure if these lines do anything?

mattxwang (changes requested):
Hi Tanaya,

I can review this PR! First, I want to say, I really like the direction that you took this navigation bar - putting the logos prominently definitely makes it easier for people to know what's going on. 

As a heads up, I know I left **a lot** of comments, so please let me know if you have any questions / want to work through these things together - it's all part of the review process! And some of these things are nitpicky, I think you've done a great job with this feature so far.

I've left some specific code comments, but here are also some general ones.

Unfortunately, the current design doesn't work well on smaller viewports, like what you'd see on a phone:

Screen Shot 2021-04-18 at 7 46 04 PM

We should definitely figure out a solution for this. One suggestion is to create a hovering menu button that appears on smaller viewports; tapping that menu can full-screen the navigation between committees, similar to the current navbar's functionality on mobile. 

FWIW, I really like how [Stripe does their mobile navigation](https://stripe.com/).

Similarly, if the viewport is really short (ex an old iPhone or Android), we have a related problem:

Screen Shot 2021-04-18 at 8 03 01 PM

You may be able to kill both birds with one stone using a smaller-viewport menu!

The last general thing: directly clicking on a committee hides the top portion of the committee image:

Screen Shot 2021-04-18 at 8 05 14 PM

This is a common problem with fixed navbars (like ours); it did scroll to the right part, but the top bit is covered by the navbar! This is a really common problem, so you're def not alone. One way to handle it is to use a negative margin + padding trick on the `:target` pseudo-class, like [the Teach LA Website](https://github.com/uclaacm/teach-la-website/blob/master/_sass/_basic.scss#L6) - just replace `70px` with the height of the navbar!

src/components/Committees/SidebarItem/MobileNavItem.js
```diff
@@ -0,0 +1,27 @@
...
+            <NavHashLink to={'/committees#ai'}>AI</NavHashLink>
+            <NavHashLink to={'/committees#hack'}>Hack</NavHashLink>
+            {/*<a href={"/committees" + `#studio`}>Studio</a>
+            <a href={"/committees" + `#icpc`}> ICPC</a>
+            <a href={"/committees" + `#design`}> Design</a>
+            <a href={"/committees" + `#cyber`}> Cyber</a>
+            <a href={"/committees" + `#teachla`}> Teach LA</a>
+            <a href={"/committees" + `#w`}> W</a>
+            <a href={"/committees" + `#ai`}> AI</a>
+            <a href={"/committees" + `#hack`}> Hack</a>*/}
```
mattxwang, line 22:
```suggestion
```

mattxwang (approved):
In general, this looks great - I think it's good to merge! Amazing work to both you and @annaguo1012, y'all are absolutely killing it with this new feature 😊 

As an aside, I imagine we'll restructure a bit of this in #138, #121, and related tickets - but we can push those changes to then!

(p.s. left one small nit on leaving a comment in the code)

mattxwang (approved):
Positive review still stands! Just make sure to squash-and-merge 😊

advaithg (approved):
Looks good to me! 

The only thing i noticed is in the case where the viewport is wide enough for the committees to show, but too short for them to be comfortable. 
This is what happens:

image of committee navigation when width is large enough for visibility but height is short

it gets squished in a weird way, and further reducing the height makes it worse. 
Although, I doubt this is a situation people actually run into - so it should be fine

The only other problem is the logos, which I assume is for another issue.

Most of the changes are in styling the navbar, and they all look good!

Great job on this PR!! It's amazing how much work you put into this and the result is really pretty!

PR #119, opened 2021-04-21
Fixes SocialMedia type prop

### LINKED ISSUE #97: SocialMedia component incorrectly uses type prop
While I'm on an issue grind...

I noticed that the `<SocialMedia>` component has an optional `type` prop. But, we render this prop directly into the `className` of the resulting icons, which means that we render a class `undefined` when the prop isn't used.

See `Home.js`; it has `.sm-icon .undefined` as its class signature:

Screen Shot 2021-02-09 at 4 03 52 PM

The short term solution is to use some sort of ternary before the render function in `<SocialMedia/>`, and render that; ex:

```jsx
const iconType = props.type ? props.type : "";

...
<button className={"sm-icon " + iconType}>...
```

Long-term though, we may want to evaluate how we implement these classes and/or how we define our components compositionally.

mattxwang:
Relatively self-explanatory.

Closes #97.

### DIFF
--- a/src/components/SocialMedia/SocialMedia.js
+++ b/src/components/SocialMedia/SocialMedia.js
@@ -1,14 +1,15 @@
 import React from 'react';
 
 function SocialMedia(props){
+    const iconType = props.type ? props.type : "";
 	return (
 		<React.Fragment>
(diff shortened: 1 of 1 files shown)


evanzhong (commented):
lgtm

evanzhong (approved):

PR #125, opened 2021-04-29
Committees page update Spring 2021

jtsujihara:
fixes #107

Issue #114, opened 2021-04-19
Stricter CI for CSS (stylelint), JS (eslint)


mattxwang:
As our codebase becomes larger and changes hands more frequently, automated style enforcing will go a long way. We can do that with **continuous integration**, where we run certain tests on our code every commit/push.

In other words, if you like automating things, this issue is for you!

The simplest CI integrations we can do (other than building, which we already do) are linting: static code analysis to make sure we don't have errors, avoid bad practice, and maintain a consistent style across the entire repository.

There are two tools we can look to loop in for now:

- [ ] [stylelint](https://stylelint.io/), a highly-configurable CSS linter
- [x] [eslint](https://eslint.org/), the defacto-standard JS linter (partially built in already)

I already wrote up a similar guide on how to integrate both of these into our workflow [for Teach LA](https://github.com/uclaacm/tla-dev-intern-training-w21/tree/main/week-6d-ci-cd-tests#cicd-in-practice-linters), so I encourage you to read through that; it should relatively walk you through how to set these up.

For posterity's sake, I suggest you:

0. make a new branch,  etc.
1. install and configure stylelint; make sure that you can run it! **do not run the autofixer yet**
2. configure an action to run stylelint on every commit. **do not run the auto fixer yet; it's okay if the action fails**
3. open up a draft PR and request a review
4. once that's approved, then, run the autofixer
5. wrap up the PR and merge it 
6. rinse and repeat 0-5 for eslint

Some things I'll note:

* for now, we can use a lax JSX/js styleguide. [Bryan from Creative Labs](https://github.com/uclaacm/website/issues/67#issuecomment-814538707) recommended [Creative Labs' config](https://github.com/UCLA-Creative-Labs/sunshine/blob/master/.eslintrc.js), which I quite enjoy. 
* eslint is already part of Create React App, so you won't need to do *as much* setup as stylelint
* standard config for stylelint is pretty good imo, you won't need to make any tweaks
* we use SASS (Syntactically Awesome Style Sheets), so you may want to install a sass plugin for stylelint as well; I included this in the [writeup](https://github.com/uclaacm/tla-dev-intern-training-w21/tree/main/week-6d-ci-cd-tests#stylelint-addendum-sass)

Implementing this then paves the way for us to add some pre-commit hooks, to automatically run the fixer - but that's for another issue 😉 . In parallel (or afterwards), we can apply this same approach for accessibility, link-checking, etc.

Supersedes #67.

mattxwang:
[...]

PR #126, opened 2021-05-02
Adds more opinionated ESLint rules, autofixes errors (CL's config)

mattxwang:
This PR adds a handful of more opinionated ESLint rules, and then autofixes (and in a few cases, manually fixes) the errors. Big shoutout to our friends at @UCLA-Creative-Labs who encouraged us to use [their config](https://github.com/UCLA-Creative-Labs/sunshine/blob/master/.eslintrc.js)!

Here's a cliff's notes of the changes we made:

* adds the `import` plugin for eslint
* style: forces single quotes, semicolons
* spacing: spacing in prop declarations, no multiple empty lines, no trailing spaces
* enforces max line-width (with exceptions)
* enforces dot notation for object fields
* disallows bitwise `|` 
* disallows `console`
* enables comma-dangling (thanks @evanzhong for bringing this up in #120) 
* manages imports: forced dep resolution, properly declared in `package.json`, strict ordering, no duplicates 

Note: we may want to hold off on merging this until #125 is in - the quote changes will need to be re-applied.

This is a follow up to #120 and is part of #114!


mattxwang:
A note: there are quite a few arguments for swinging quotes in either direction - I'm open to changing it to double!

src/serviceWorker.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
evanzhong, line 1: Assuming this is intended

evanzhong, line 1: assuming this is intended

evanzhong (approved):
Scanned over all diffs.

Main things I see from lint
- whitespace changes
- double quotes => single quotes
- added trailing commas
- added missing semi-colons
- line breaks due to maximum line length 
- alphabetization of imports

Otherwise lgtm

src/serviceWorker.js
mattxwang: Yup, we don't enable the service worker anyways but I don't want to interfere w/ the boilerplate code

mattxwang:
Actually if you don't mind, @bryanpan342 is there a reason why y'all chose to do single quotes over double quotes? Might as well ask the source haha

BryanPan342:
> Actually if you don't mind, @BryanPan342 is there a reason why y'all chose to do single quotes over double quotes? Might as well ask the source haha

I think it's a personal preference. I find single quotes a lot more readable, esp when it comes to empty strings! There is an argument that it's annoying for apostrophes, but I think at the end of the day, if you are using React IDEALLY you have a CMS populating the content for your static site. And in those cases, the apostrophe problem doesn't hold anymore.

When I look at most libraries they use single quotes. So I think my coding style has revolved around those libraries. Regardless of what you pick, as long as you enforce the same style I think it's fine.

mattxwang:
Tysm for the v detailed explanation, we've gotta put a few CL shoutout easter eggs in the website 😊

.eslintrc.js
```diff
@@ -2,29 +2,107 @@ module.exports = {
...
-    node: true
+    node: true,
   },
   extends: [
-    'eslint:recommended',
-    'plugin:react/recommended',
-    'plugin:jest/recommended',
+    "eslint:recommended",
+    "plugin:react/recommended",
+    "plugin:jest/recommended",
```
advaithg, line 10: Is there a reason this is the only file with double quotes used instead of single quotes? Every other file seems to have been changed to use single quotes - whereas this one has been changed to use double quotes.

advaithg: I think the rest of the files changed to single quotes due to the autofix, but does that convention not apply to this file for a particular reason?

mattxwang: That's a good point! Currently, the path for this file isn't hit by the glob for `npm run lint-js-fix`, lemme change this one manually.

advaithg (approved):
I looked through the files and the other sources, and I didn't notice anything beyond what I already pointed out and what Evan listed out.

Looks good to me!

mattxwang:
For posterity, resolved the merge conflict:

* took the new dep bump in `package.json` from #132 
* took every incoming change (i.e. jeremy's stuff in #125) in `committee.js`, then ran `npm run lint-js-fix`

PR #135, opened 2021-05-04
Manually change links on committees page to use <a> tag instead of anchorme

### LINKED ISSUE #93: Use of "here" for committee links is inaccessible
On the [committees page](https://acm.cs.ucla.edu/committees), we use `react-anchorme` to automatically insert `<a>` tags with `"here"` as the target text. This is not a bad solution, but is not particularly accessible (nor is it good for SEO); `"here"` isn't really descriptive of what the link text represents, and is particularly troublesome for screenreaders. 

I think there are a few options we have:

* directly embed the `<a>` into the `.json` file that renders the component; given that this is a JSX object, this is totally fine, and probably the easiest way to do it.
* run the text input through some sort of markdown generator that spits out HTML. requires input sanitization
* use an icon to represent the actual link, and then use an `aria-*` attribute to describe what the link actually points to 
* something else I may be missing!

Let me know what y'all think!

changyang-liu:
Resolves #93. Change all descriptions with links to be <p> paragraphs instead of strings, and change the anchorme links to be <a> tags. Also remove the anchorme dependency.

Note: the text of the anchor tags was made at my own discretion, and consists of either a URL or descriptive words, so please let me know if you find any of them to be unsuitable (e.g. if the descriptions are not clear enough).


mattxwang (changes requested):
Hey @changyang-liu, in general the changes work well! I just have one question - is there a reason why you've converted all the single quotes to double quotes? It ends up polluting the git history (and makes the commit diff quite large).

If we revert that, then I think this PR will be good to merge!

changyang-liu:
> Hey @changyang-liu, in general the changes work well! I just have one question - is there a reason why you've converted all the single quotes to double quotes? It ends up polluting the git history (and makes the commit diff quite large).
> 
> If we revert that, then I think this PR will be good to merge!

Oops, my prettier extension did that automatically without me noticing. I'll change it back right now

mattxwang (approved):
Looks good to merge on my end!

P.S. the wide majority of these changes are still Prettier whitespace/quote changes; that being said, we're doing something similar in #126, so I'll let them in this time! In the future, I'd recommend disabling your default config and instead reading in the provided `.eslintrc.js` file (not sure how to do that with your extension).

mattxwang:
As a heads up @changyang-liu, there are a few merge conflicts (I think they mostly stem from some prettier whitespace changes on your end); how I'd approach it is:

* for the `CommitteeEventCard.js`, reset the whitespace and then re-delete the `<AnchorMe />` component
* for the `committees.js`
  * if the conflict is *just* spacing/quotes or something, take your change (i.e. the one at `feature/...`), then run `npm run lint-js-fix` to reapply the spacing changes
  * if the conflict is more than just that (i.e. a content change), that's likely something from Jeremy's PR #125 - I would then take his change (i.e. the one at `master`), then re-apply the link change (sorry!)

Let me know if that makes sense, or if you want to hop on a call and walk through this together!

mattxwang (approved):
Diff looks much cleaner! LGTM!

mattxwang:
> Which do you normally use in this repo?

Great q! Normally, I would squash and merge - I only rebase when the commit history itself is relevant to the PR (in this case, I would say no).

Issue #123, opened 2021-04-26
Remove "winter gm" page


evanzhong:
Maintaining code is just as (if not more) important than writing new code. Removing legacy code, or code that's simply outdated is a difficult task, and there are many opinions on the "right" way to do so.

Due to virtual operations, I previously created a [landing page](https://acm.cs.ucla.edu/gm) for our virtual general meetings. We're now more than a quarter removed from our last general meeting. This ticket tasks you with removing the landing page or somehow archiving the code related to the page.

Things to consider
---
- What new component(s) were defined for this feature, what's reusable, what's not?
    - On that vein, should we do a "full rollback" (removing all code relation to the GM landing page)?
    - Or should we somehow further generalize and component-ize the code, document its existence, and move it to an archival part of the code base.
- What dependencies did this feature rely on? How do we roll this back? Do we need to roll it back?
    - How does your choice to the previous bullets affect your choice here about whether to roll back installed npm packages?
- How did this feature affect navigation? How can you find all references to this component / page and remove references to it?

Helpful resources
--
- Looking at [past Pull Requests](https://github.com/uclaacm/website/pulls?q=is%3Apr+gm)
- Exploring the code in `src/components/Events/GM/`

advaithg:
I think a good solution to this would be to just archive the page and maintain the code/file while removing all references to it from the main web page. That way we can reuse the exact same page for later GMs simply by un-commenting references to it. Would that make sense or is there a problem with that idea?

evanzhong:
> I think a good solution to this would be to just archive the page and maintain the code/file while removing all references to it from the main web page. That way we can reuse the exact same page for later GMs simply by un-commenting references to it and making some changes in content. Would that make sense or is there a problem with that idea?

This solution can definitely work! Some additional things to consider if you choose this approach
- How do we make it clear that a component or set of files is "archived"?
- How do we document how to "revive" a page if / when a future dev chooses to do so?
- What should we do about the assets (i.e. images, references to slides etc) that this component relies on.
    - How do we generalize these, since currently all the images are specific to a particular GM (i.e. Winter 2021)?

Would also love to hear @mattxwang's take

mattxwang:
@advaithg I like your initial solution, and @evanzhong I think the points you bring up are pretty valid!

One approach we can consider:

* what other orgs often do is create a route called `archive` and place past events there, for posterity. For example, we could have `uclaacm.com/archive/w21-gm` be a route for the existing component.
[...]

PR #141, opened 2021-05-08
Remove Winter GM Page

advaithg:
Addresses issue #123

Changes made:
1. Removed references to GM from navbar
2. Replaced most common data using placeholders - placed data in **config/gmData.js**
3. Imported data file and **replaced hardcoded information** with placeholders
4. Added **comments** detailing gmData file and how to add back to navbar

Did not remove route from router - page is still accessible through '**/gm** '


src/components/Events/GM/GM.js
```diff
@@ -130,40 +95,38 @@ function GMLiveProgram() {
 					</div>
 				</div>
 				<div id="video-wrapper">
-					<iframe title="ACM Winter GM 2021 recording" src="https://www.youtube.com/embed/cleSPlz2U1Q" frameBorder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowFullScreen></iframe>
+					<iframe title={'ACM ' + gmData.date.quarter + ' GM ' + gmData.date.year +' recording'} src={gmData.links.recording} frameBorder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowFullScreen></iframe>
```
evanzhong, line 98: you can similarly try string templating here

src/components/Events/GM/GM.js
```diff
@@ -130,40 +95,38 @@ function GMLiveProgram() {
...
 					</div>
 					<div className="row">
-						<p className="timestamps">7:10 p.m. PST</p>
+						<p className="timestamps">{gmData.times.social + ' ' + gmData.times.timezone}</p>
 						<div className="row-description-wrapper">
 							<h3 className="row-description-heading">Tabling and Social</h3>
-							<p className="row-description-subheading">Join us in our official <a href="https://bit.ly/ACMdiscord"  target="_blank" rel="noopener noreferrer" id="acm-discord-invite-cta">ACM Discord</a> for Committee tabling and socials</p>
-							<p>Daniel Jaffe, Evan Zhong</p>
+							<p className="row-description-subheading">Join us in our official <a href={gmData.links.discord}  target="_blank" rel="noopener noreferrer" id="acm-discord-invite-cta">ACM Discord</a> for Committee tabling and socials</p>
+							{/* <p>Daniel Jaffe, Evan Zhong</p> */}
```
evanzhong, line 129: I would actually remove this line entirely, or generalize it rather than commenting it out

src/components/Events/GM/GM.js
```diff
@@ -198,10 +161,10 @@ function GM_COUNTDOWN(props) {
 					<div className="countdown-labels">{props.secondString}</div>
 				</div>
 			</div>
-			<h3 id="gm-zoom-release">Zoom link, slides, and other resources will be provided on Thursday, January 7th, 2021 at 6:30 p.m. PST</h3>
-			<p id="posted-to-youtube">The Zoom portion of Winter General Meeting 2021 will be recorded and posted to our <a href="https://www.youtube.com/c/acmucla" target="_blank" rel="noopener noreferrer">YouTube channel</a>.</p>
-			<p id="cant-wait">Can&rsquo;t wait to join in on the fun? Jump in right now with our <a href="https://linktr.ee/acmucla" target="_blank" rel="noopener noreferrer">Linktree</a> or hop into the official <a href="https://bit.ly/ACMdiscord" target="_blank" rel="noopener noreferrer">Discord</a> and say hello!</p>
-			<img id="gm-graphic" src={process.env.PUBLIC_URL + '/images/Winter_GM_2021_graphic.png'} alt="Winter General Meeting 2021 graphic" />
+			<h3 id="gm-zoom-release">Zoom link, slides, and other resources will be provided on {GM_START_TIME.toLocaleString('en-US', {weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' })} at {gmData.startTime} {gmData.times.timezone}</h3>
```
evanzhong, line 164: Not a huge issue, but you can consider moving the function call to `toLocaleString` into a separate helper function. (i.e. this bit `GM_START_TIME.toLocaleString('en-US', {weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' })`)

In my personal experience, moving as much complicated logic (or javascript expressions/statements that take a moment to read) to dedicated functions helps with code separation. 

In brief, you want to keep as much of our "template" (html) separate from our "controlling logic" that actually generates the dynamic content.

src/config/gmData.js
```diff
@@ -0,0 +1,109 @@
+export default {
```
evanzhong, line 1: Over data structure looks great!

I'm now realizing that these data files would be great places for Typescript, so we can enforce more strict compile time rules about formatting (i.e. the comment you wrote below about `month` and `day`. Would love to hear @mattxwang's take on this.

Perhaps out of scope for this PR, but you could also consider adding some smoke tests for the data. (i.e. regex checking length of strings, formatting etc).

src/config/gmData.js
```diff
@@ -0,0 +1,109 @@
...
+        eventName: 'Event 1',
+        speakers: 'Speakers',
+    },
+    {
+        logo: '/images/logo.png',
+        displayName: 'ACM',
+        eventName: 'Event 2',
+        speakers: 'Speakers',
+    },
+];
```
evanzhong, line 109: I'm not sure what our exact procedure for our codebase is, but my preference is to define each data structure as just a constant variable, then consolidating all export statements at the end.

Something along the lines of:
```suggestion
];

export {
    boardEvents,
    committeeData,
    //your default obj
}
```

Alternatively you could just wrap everything as one big ol' object, then export the large wrapper object as a default.

mattxwang:
As a quick comment, I just merged #147 which slightly changes the `eslint` behaviour (it does an extra autofix step), when you have a chance @advaithg can you merge from `master` and rerun + push?

src/config/gmData.js
advaithg: I ended up deciding against that since I wasn't sure what the error message or error handling situation for that should be - since right now giving it an invalid input makes date NAN - which makes it obvious along with the comment what the error is.

Do you think I should still do it? If so what kind of error message would you say make sense?

evanzhong: I think it's probably more work than its worth and I'd rather ship this PR as is for the following reasons
- The data model is pretty well documented in comments and in this PR
- This file will likely see pretty light use, given we only have 2 GMs every year

Final call is up to you.

advaithg: This makes sense! 
Once it's done being reviewed I'll merge it into main then.

evanzhong (approved):
lgtm!

annaguo1012 (approved):
LGTM! I think Evan covered everything. I really like how this turned out and appreciate the helpful comments for future devs! :)

PR #147, opened 2021-05-12
Adds `eol-last` rule, autofixes relevant files

mattxwang:
Short PR, but @evanzhong pointed out I forgot to enforce newline at the end of the file in #126, so this PR does that and then autofixes the relevant files.


evanzhong (approved):
lgtm!

PR #150, opened 2021-05-13
Added Stylelint Configuration

advaithg:
Addresses issue #114 

Changes made
1. Installed required dependencies for stylelint and SCSS
2. Created .stylelintrc.json file and set up default configuration
3. Added as CI action in .github/workflows/lint.yml
4. Updated scripts in package.json to run lint-css and lint-css-fix

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,5 +25,4 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm install
-    - run: npm run lint-js
-    - run: npm run lint-css
+    - run: npm run lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,3 +26,4 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - run: npm install
     - run: npm run lint-js
+    - run: npm run lint-css
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -5,8 +5,11 @@
     ],
     "ignoreFiles": ["dist/*", "node_modules/*"],
     "rules": {
-      "at-rule-no-unknown": null,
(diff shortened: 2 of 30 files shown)


advaithg:
It fails the lint test since I did not run lint-css-fix yet, will do once this is approved!

mattxwang (changes requested):
Hey Advaith,

Exciting stuff! I think we're on the right track here 😊  just some things I'll make note of:

* for `color-hex-case`, the default for the rule is to enforce lowercase; but, looking at the [Design Styleguide](https://design.uclaacm.com/committees/acm/), it seems like ACM Design prefers uppercase. Do you mind changing the rule?
*  after running the autofixer, did you see any false positives that seem unreasonable? given the nature of our codebase (and also the SCSS plugin), there may be some things we'd have to look into. Alt., if almost all the errors are autofixable, then we're in a good spot!
* I'm curious, if we use the [recommended config for stylelint](https://github.com/stylelint/stylelint-config-recommended), do we run into any compat issues? it's a bit stricter (which I personally prefer)
* this is very minor, but do you mind making a `lint` command for `npm` that runs both `lint-js` and `lint-css`? makes life easier for developers 😊 

I think we're on a good track to merge this soon! Once we wrap up our investigations, we can commit the fixed files, and then merge the PR after some reviews!

mattxwang (approved):
LGTM, thanks for addressing the changes! Think we're in a good spot 😊

advaithg:
This changes only 11 files, so I decided to check with the standard stylelint config to ensure the recommended config (which is what i used) is working fine.
Turns out that is not the case, since standard shows a bunch of autofixable issues - which the current recommended stylelint config does not seem to be catching
To fix before merge!

mattxwang (approved):
Approving review still stands, LGTM!

dtjanaka (approved):
LGTM

Issue #131, opened 2021-05-04
Flattening/restructuring the component directory


mattxwang:
This is an issue that I will probably handle, but some of these component paths are a bit absurd, ex

![Screen Shot 2021-05-03 at 5 48 56 PM](https://user-images.githubusercontent.com/14893287/116949684-da43fe80-ac37-11eb-93ea-f1f3371f75da.png)

Some goals for me (as a short-term refactor):

* can we differentiate between pages and non-pages? is this too destructive?
* can we localize some shared components into one folder
* if a component is only ever a child of another component and is completely static, can we instead put it inside another class file?

mattxwang:
Blocked (and/or will be resolved) by SSG rework!

PR #170, opened 2021-06-11
Flattening "committees" page components

mattxwang:
Hello there! This is a convenience PR to knock out some refactoring (#131) before we make a big site migration (which would involve changing a bunch of components). Basically, this PR does three things:

* moves tiny components with no state into their parent component, to reduce nesting; involves some renaming
* collates styles together 
* adjusts some CSS to make the above two changes no-ops on visuals

Overall, the impact on the end user should be *nothing* - this PR just aims to improve the developer experience. I've tested locally on my end side-by-side and I see no issues, but a second look would be great! 

Two quick notes:

* left some TODOs on weird JS things I saw
* our CSS has way too much overlap - hopefully this is something we can resolve with Gatsby/Next, as we can build custom bundles for each page. we may also want to rethink how a lot of our current CSS is written, there's way too much namespace overlap and specificity mixing


advaithg (approved):
Looks good to me!

Deploy preview works fine
Comments make sense to me - they're things we should do when working on the transition

Major changes are all in removing unnecessary files with minimal code imported only in one place and putting the code in the places where it's imported
The most obvious changes are in the Committee Sections, with more changes in the banner, sidebar and navbar
Styling is also more centralized now

Probably can still make it cleaner if possible, considering the repeated CSS and still somewhat convoluted paths in some places (like `src/components/Committees/CommitteeSections/CommitteeSectionIntro/CommitteeSectionIntro.js`)

PR #173, opened 2021-06-16
Replaces Events page content with a "we're on pause" disclaimer

### LINKED ISSUE #172: Put up an interim "events are on pause" page
Our events page looks pretty empty since we're generally not running stuff over the summer. We should put something up on the page that indicates that we're still an active club that just takes breaks during the summer! Some ideas:

* a simple "we're on break, but catch us soon"
* past recordings / links to workshops
* general outline of what's coming up in fall
* etc.

mattxwang:
Self-explanatory. Closes #172.


advaithg (approved):
Looks good to me!

Changes I noticed were adding a bit of styling, commenting out the calendar, and mentioning we're on pause, and they look fine!

PR #174, opened 2021-06-16
Consolidating global styles, remove unused classes part 1: fonts (mostly)

mattxwang:
The general changes:

* splits up `Base.scss` into two files, `variables.scss` and `globals.scss` to split apart variables and global styling
* enforces header font styling (font family, color) on a global level, and removes (element-selector) individualized classes
* uses `$navbar-width` more frequently instead of magic numbers
* removes unused CSS (inspected w/ dev tools)
* adjusts some tags to use semantic HTML properly
* renames `center` to `.text-center`, moves to `globals.scss` for better visibility; de-dupes from other styles


advaithg (approved):
Lgtm!

The changes look like mostly moving around CSS and consolidating it better, getting rid of some repetitive styling.

PR #175, opened 2021-06-16
Moves Tech Gala Tag to ProjectCard, cleans up CSS

mattxwang:
Part of #131!

(otherwise, self-explanatory)

Note: we should eventually de-dupe the images with the `.intro-row` class!


evanzhong (approved):
lgtm, nice catch with the alt text as well

PR #176, opened 2021-06-18
Fix image alignment on `About.js`

Alvin-N:


mattxwang (approved):
Great catch - I must've accidentally added this. @advaithg do you want to give one more approving review?

PR #177, opened 2021-06-18
fix: cleans up committees sidebar onScroll with `useEffect`

mattxwang:
This is a short and sweet PR that resolves a small regression introduced in #108 - the `window.onScroll` is created as a side effect on render, but is never properly cleaned up. This resolves that by using the `useEffect` hook and returning a cleanup function to the hook.

As a side note: I make the re-render dependency list empty since I don't think a change in `bottom` should trigger a cleanup.


advaithg (approved):
Lgtm!

Fixes the cleanup issue we discussed!

Issue #116, opened 2021-04-19
Populate page metadata, social media metadata


mattxwang:
One of the follies of Single Page Applications (SPAs) is that you need to do a bit of extra work to make the website play nicely with web crawlers (ex google's preview) and social media embeds (ex links from Facebook, Twitter, Discord).

Luckily, this is a relatively solved problem. Given the (so far) small-ish number of pages on our site, this is something that we can do mostly manually, with some minor automation if needed.

There are several disjoint things we need to consider:

1. Properly filling out `<meta>` tags in the document head.  One defacto standard is [`react-helmet`](https://github.com/nfl/react-helmet).
2. Properly dealing with [Open Graph](https://ogp.me/) metadata, commonly used for social media embeds (and developed by Facebook). This is quite hard, since these crawlers (unlike Google) tend to not run JS bundles.
3. Customizing for certain networks, such as [Twitter Cards](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/abouts-cards) - this is a nice-to-have, and not a requirement for this ticket

## Social Media Optimization with JavaScript

Open Graph is by far the trickiest part of this situation, since SPAs by nature don't work properly without JavaScript, and most crawlers don't run JS when they crawl a site. To remedy this, there are a few options; here they are in orders of complexity:

1. adding a default OG metadata to the public `index.html`. Not great, but better than what we have right now.
2. investigate a hacky solution like [`react-snap`](https://github.com/stereobooster/react-snap), which prerenders + crawls our webapp to create HTML pages for crawlers.
3. use one of the frameworks designed to solve this problem / take advantage of Server Side Rendering (SSR): [Next.js](https://nextjs.org/) and [Gatsby](https://www.gatsbyjs.com/) are likely the two best contenders.

I personally would *absolutely love* to do the last option, though it has its own ramifications around maintainability (we need to teach people Gatsby/Next) and the likely insane amount of merge conflicts this would generate with other PRs.

Also would encourage you to ask around / do your own research around this problem - I (Matt) am definitely not as in the loop as other web developers on the perfect way to resolve this problem. Some possible ACM-adjacent people to discuss this with:

* ACM Hack is a big fan of Gatsby, they use it for [their website](https://github.com/uclaacm/hack.uclaacm.com) and [HOTH](https://github.com/uclaacm/hoth.uclaacm.com)
* Creative Labs uses Next frequently, including for [their website](https://github.com/UCLA-Creative-Labs/sunshine)

[...]

Issue #110, opened 2021-04-15
Consider auto-generating sitemap


mattxwang:
I was poking around on things on GitHub (as one does), and I think it'd be cool to auto-generate a sitemap. 

We should be able to do it programmatically; since we're an SPA right now, we probably want to use something like [react-router-sitemap](https://www.npmjs.com/package/react-router-sitemap). Netlify also has an [interesting plugin](https://github.com/netlify-labs/netlify-plugin-sitemap), but I imagine it needs some tweaking to work with CRA/react-router out of the box.

mattxwang:
Blocked by SSG rework!

mattxwang:
Just installed the Netlify plugin, so we can close this issue. Works as intended; see https://www.uclaacm.com/sitemap.xml

Issue #115, opened 2021-04-19
Long-term: Reworked Events Page


mattxwang:
Likely our most requested feature is to significantly amp-up our [Events page](https://www.uclaacm.com/events), which is currently just a Google Calendar embed. Turns out, making good events pages is ... hard.

Three limiting concerns that drive a more aggressive rework:

1. Google Calendar is a limited service: for example, we cannot embed images, and the views are not easy to customize.
2. Facebook is used less and less by incoming classes, and thus we need a way to convey event information without requiring Facebook use.
3. There's currently no central authority where general members can find all of ACM's events for a given week. Google calendar has served as a patchwork solution, but has drawbacks listed above. With a revamped events page on our website, we can offer our community this resource.

Given the large scope of this task, I want to split it up into three (mostly) disjoint things:

1. **Data**: how should we store the data for each event? do we host it internally, or use another service and hit an API? how does the membership portal play into this? How do we balance that committees will likely have different needs?
2. **Design**: how should the page look? can we make it responsive (given that a significant amount of traffic is mobile)?
3. **Interface**: can we make it easy for users to see all the relevant event information, without being overwhelmed? can we make it easy to filter for events based off of committee, type, etc.? what about people creating an event?

I want to stress that this is a **large ticket**, and it is:

* fine to be lost, since this document is **long**
* encouraged to split this up upon multiple PRs
* totally okay to ask for help!

## The Data

First, we need to decide what data is needed for each event. Examples include:

* "general" info: name of event, who's holding it, when it is
* online and/or in-person location
* blurb describing the event in detail
* relevant resources (ex GitHub, slides, recordings)
* images?

Next, we need to decide how to *store* this information. There are a few options; however, if possible, I would like to have a **single source of truth**. Here are some, off the top of my head:

* store the data in the repository. Marketing already has a spreadsheet with most relevant event information quarter-on-quarter; we can write a script to convert the CSV -> JSON, then render out of the JSON. **this is the lowest friction option**, in my opinion.
* use the **membership portal** as a single source of truth; this was the plan in #9, though I'm personally unsure of how feasible this is.
* write our own service that provides event information as an API; then, rely on this service for the membership portal AND the static website
* use Google Calendar as a "headless CMS (content management system)" for our data
* somehow use Facebook (unclear on how to do this without violating Terms of Service)
* some other approach that I haven't thought about!

[...]

PR #178, opened 2021-06-18
Switch to Next.js

### LINKED ISSUE #131: Flattening/restructuring the component directory

### LINKED ISSUE #162: Scroll location on page change
When switching from one page on the website to another via the navbar or the logo button, the page opens at the same corresponding section (at the same location) instead of opening at the top

i.e. if you scroll to the bottom in the committees page and click the navbar or the main logo it will take you to the bottom of the corresponding page you clicked on, instead of to the top (which would be better behavior)

According to @mattxwang this probably has to do with some `e.preventDefaultScroll` shenanigans, and should be auto resolved on switching to next.js / gatsby - so just putting this up for posterity / as a reminder!

mattxwang:
This PR does the meat of our transition to Next.js. As a heads up, the files changed/diff is absolutely nasty; I imagine that there will be little bits and pieces that we'll have to address down the line.

## Summary

We transition the site to be Next SSG'd.

### CRA-to-Next transition

* swaps out toolchain to use Next
  * installs `next`, and relevant `eslint` plugins
  * removes `react-scripts`, `react-router-dom`, `react-router-hash-link`, and `@testing-library` equivalents
  * changes `npm` scripts to use `next` defaults
  * swaps out `node-sass` for `sass` (the reference dart implementation)
* uses Next's file-based routing
  * currently segments into six `pages`: `about`, `committee`, `events`, `index` (previously `Home`), `sponsors`, and `techgala`; no paths are broken
  * moves the rest of components (including `GM`) into `components`
* restructures SASS and CSS
  * moves all styles into `styles` folder, mirroring `pages` and `components` structure
  * collates a few styles
  * imports Font Awesome into the next bundle, instead of loading into the `index.html` entrypoint
* uses `next/image` when possible
  * in some cases, we import the image out of the `public` folder; this automatically sets `width` and `height`
  * in some other cases, manually sets `width` and `height` (typically for UI elements) or uses `layout="fill"` and retools CSS to properly deal with it
  * **in some cases, maintains ``. we were unable to fully transition over to `<Image />` (see below conversation)**
* for ease-of-transition, removes all tests (`jest`, `enzyme`, related things)

### Things that will need to be addressed later

* this PR explicitly doesn't address `.module` SASS, but we should do that in a follow-up
* ~there should be something we can do about fonts that doesn't involve loading it in the `<Head>` - let's look into that!~ possibly see @BryanPan342's comment below
* SEO, using `<Head>` properly, etc; since this isn't a regression, let's table adding this to each page in a separate PR
* more ergonomic image inputs - maybe through an `index.js` that collates together images, thanks @dtjanaka for the suggestion!
* remove `postcss@8` as a dev dependency, but still resolve this error: https://github.com/vercel/next.js/issues/24882
* refactor the `GM` component to derive all of its information from state; then, make a new wrapper page that passes in the relevant information from data files, manual configuration, etc.
* document the new CI/CD pipeline, document Next quirks (ex `next/image`)
* resolve issues with `next/image`

### Misc. code quality changes

* removes the `<Button>` component entirely, and changes the `button` selector to instead use the `.button` class.
  * this is because the component itself didn't do anything (it just passed on `props` to a `<button>`) and also arbitrarily restricted the semantic element
* changes CSS for the officers section of the about page to use flexbox instead of nested selectors + absolute/inline-based positioning
* changes the officers section text to say "Leadership" <- this is easily changeable
* fixes bug where `.page-content p` bleeds into the Committees page events arbitrarily
* fixes incorrect use of `ul` and `li` in `<Navbar>`
* slashes a non-trivial amount of completely unused CSS
* de-dupes `CommitteeSectionIntro`'s double-images for committee motifs

### Issues that still need to be addressed:

- [x] switching over to `next/image` to resolve build failures
- [x] properly using `next/image` to resolve layout issues - see `TODO` comments
- [x] font awesome icons are not sized properly
- [x] strip out the `Button` component, as it's not useful and causes issues with `Link`
- [x] fix the Navbar button (general_
- [x] fix the Navbar membership portal button on mobile
- [x] ~~this error: https://github.com/vercel/next.js/issues/24882~~ - stopgap is to have `postcss@8` as a devdep; moved to "deal with later"
- [x] resolve netlify build error (very simple, just change output folder)
- [x] ~~deal with the `GM` component -> this should probably be a component, and then have a dummy page wrap it for every new GM. deals with longevity very well~~ - moved to "deal with later"
- [x] fix about page first image diamond (now not centered properly)
- [x] fix image borders not appearing properly on `next/image` components
- [x] fix committee page nav images not being properly sized
- [x] the favicon is bugged

This PR also does some favours for us: it closes #162, (reasonably) closes #131 (for now). It also unblocks #110 and #116!


BryanPan342:
@mattxwang lmk if u need help with the next.js migration!

Even tho next is awesome, sometimes the bugs are opaque AF and takes a bit of digging to fix!

mattxwang:
> @mattxwang lmk if u need help with the next.js migration!
> 
> Even tho next is awesome, sometimes the bugs are opaque AF and takes a bit of digging to fix!

Thanks, appreciate it! I might pick your brain once I resolve all the low hanging fruit. Gotta say, the image conversion has been the largest pain point thus far, though it's also a huge code smell with our website in general!

BryanPan342:
For fonts, usually we import them into our variables or global scss file!

https://github.com/UCLA-Creative-Labs/sunshine/blob/master/styles/_variables.module.scss

pages/_document.js
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+
+class AppDocument extends Document {
+  render() {
+    return (
+      <Html lang="en">
+        <Head>
+          <meta charSet="utf-8" />
+          <meta name="viewport" content="width=device-width, initial-scale=1" />
+          <link href={'/public/images/logo.png'} rel="icon" type="image/png" />
```
BryanPan342, line 11: You can link to `/images/logo.png`

Next.js takes ur public folder and treats it like root!

BryanPan342 (commented):
Why are u using document? What part of the html/body tag do you really need to target?

pages/_document.js
BryanPan342: Yeah I think anything with a `public 
_url` you should be able to just remove! Also helps you from having to deal with that during production, which definitely won’t work with the base webpack plugins

mattxwang: Great catch! I'll resolve;

> Why are u using document? What part of the html/body tag do you really need to target?

The first-party next eslint plugin [throws an error](https://nextjs.org/docs/messages/no-page-custom-font) when you try to import a font (or in general, anything that should span every single page) in `_app.js`. Seems like you're supposed to slap all of that stuff in `_document.js`, but I could be totally off base here; still new to things!

BryanPan342: Awww I see.. for fonts check out the comment I made above.. it should remove that necessity! 

As for importing things for every page, what I like to do (and what I have seen elsewhere) is basically to have a Layout component (which you already have) that takes care of all your headers!

https://github.com/UCLA-Creative-Labs/tears/blob/main/components/Layout.tsx

You can even create different Layout types and have those be utilized throughout ur app!

advaithg (commented):
I fixed the issues with the weird sizing of Next Images by making the attributes of the Image element `layout: "fill"` with `objectFit:"contain"` and wrapping them in a div which had the size we needed along with `position:relative`

This made sense cause the images on the website mostly had a fixed height but varying width, and Next didn't let me do that - so I had to do that in an outer div. I couldn't find a way to tell Next to "use default width"

Alignment was done using the `objectPosition` attribute

mattxwang:
@BryanPan342, I'm curious do y'all host CL stuff on Vercel? Running into some issues with using `next/image` on Netlify (you can peep [the deploy preview](https://deploy-preview-178--jovial-pasteur-581b4a.netlify.app/) or check out the discussion https://github.com/vercel/next.js/issues/18356 or https://github.com/vercel/next.js/discussions/19065 for some context). No worries at all on timing though, you're doing a huge help for us already!

advaithg (commented):
The current issue is that images fail to load - not sure exactly why.

@BryanPan342 do you know of a way to possibly fix this? (if you've experienced it before)

The most plausible reason we found is [this issue](https://github.com/vercel/next.js/issues/18356), but the only ways to fix it seem to be either switching to [next-optimized-images](https://github.com/cyrilwanner/next-optimized-images) or to vercel instead of using netlify.

mattxwang:
> For fonts, usually we import them into our variables or global scss file!
> 
> https://github.com/UCLA-Creative-Labs/sunshine/blob/master/styles/_variables.module.scss

Never got back to this; I'm curious, does this delay the font load time? Since then we'll have to hit the CSS file first (fetch + parse), and then load the font itself (fetch + parse)?

BryanPan342:
> Never got back to this; I'm curious, does this delay the font load time? Since then we'll have to hit the CSS file first (fetch + parse), and then load the font itself (fetch + parse)?

it does a little bit (I only really notice it when I'm really laggy), but I mean, it will load with the rest of your styling so I don't see a problem with it TBH

BryanPan342:
> @BryanPan342, I'm curious do y'all host CL stuff on Vercel or Netlify? Running into some issues with using `next/image` on Netlify (you can peep [the deploy preview](https://deploy-preview-178--jovial-pasteur-581b4a.netlify.app/) or check out the discussion [vercel/next.js#18356](https://github.com/vercel/next.js/issues/18356) or [vercel/next.js#19065](https://github.com/vercel/next.js/discussions/19065) for some context). No worries at all on timing though, you're doing a huge help for us already!

we host everything on netlify! I haven't run into many problems with next/image BUT to be honest, I prefer using just the regular `image` tag! I find `next/image` abstracts a little too much for me personally.

mattxwang:
> we host everything on netlify! I haven't run into many problems with next/image BUT to be honest, I prefer using just the regular `image` tag! I find `next/image` abstracts a little too much for me personally.

Hm okay thanks for the thought - I think we'll switch back to `` for now (since it moves gracefully), and try our hand at `<Image />` some other time (I also saw that there's some movement with the Essential Next.js plugin).

components/GM.js
```diff
@@ -0,0 +1,177 @@
+// import React from 'react';
```
dtjanaka, line 1: Why is this file completely commented out?

components/TechGala/ProjectCard.js
```diff
@@ -1,11 +1,12 @@
+import Image from 'next/image';
 import React from 'react';
 
 function ProjectCard(props) {
     const {image, title, winner, summary, names, url, category} = props.project;
     return (
         <div className="intro-row">
             <div className="show-mobile">
-                <img src={image} alt={`showcase image for ${title}`}/>
+                <Image src={image} alt={`showcase image for ${title}`} height={280} width={370} />
```
dtjanaka, line 9: Maybe a dumb question, but it seems like height and width are now required props for images; what are the implications for responsiveness?

data/sponsors.js
```diff
@@ -0,0 +1,69 @@
+import blizzard from '../public/images/sponsors/blizzard.png';
```
dtjanaka, line 1: maybe consider making an index.js file for the images so u can just do `import {blizzard, citadel, deloitee, etc.} from '../public/images/sponsors/';`

dtjanaka (approved):
nice

components/GM.js
advaithg: I think GM needs to be refactored into an actual component in a dummy page, will probably be another issue!
@mattxwang should have more info

components/TechGala/ProjectCard.js
advaithg: The images automatically readjust themselves based on the size of the page, iirc. The height and width props are for the general case.

components/GM.js
mattxwang: Yeah, just to follow up - since the page isn't live, didn't want to introduce more complexity into the change. Eventually, we should adjust this too.

A quick thought - I think the page for a GM (ex `gm-w21`) should wrap the existing `GM` component, and simply pass in props to the component. That way, the `GM` component itself doesn't need to rely on external data sources, which may be a better separation of concerns.

data/sponsors.js
mattxwang: Great idea, let's add it to the backlog 😄

components/TechGala/ProjectCard.js
mattxwang: Yeah, I think moving forward @advaithg is going to be the SME for `next/image` - you may want to discuss how the component works (ex specified with/height versus `layout="fill"`, what's happening under the hood) with the team some time.

components/Layout.js
```diff
@@ -0,0 +1,19 @@
...
+      <div className="app-container">
+        <Navbar />
+        <div className="main-container">
+          {props.children}
+        </div>
+        <Footer />
+      </div>
+    </>
+  );
+}
```
evanzhong, line 17: This is so clean, really nice

components/Navbar.js
```diff
@@ -0,0 +1,107 @@
...
+						<a id="nav-title" className="nav-section left">
+							<img src={'/images/acm_wordmark&logo.svg'} id="acm-logo" alt="Acm at UCLA Logo"></img>
+							{/* TODO: use next image without breaking deploy
+							  <Image
+								src={'/images/acm_wordmark&logo.svg'}
+								width="180px"
+								height="60px"
+								alt="AcM at UCLA logo"
+								priority={true}
+							/>  */}
```
evanzhong, line 67: Similarly, remove if possible before merging

components/Navbar.js
```diff
@@ -0,0 +1,107 @@
+// import Image from 'next/image';
+import Link from 'next/link';
+import React from 'react';
+// import gmData from '../../config/gmData';
+
+//to allow GM page to be visible on the navbar and accessible
+//uncomment the above import line "import gmData from '../../config/gmData';" to get quarter from the gmData page
+//uncomment the two lines saying  "/* <Link href="/gm"><li>{gmData.date.quarter} gm</li></Link> */"
```
evanzhong, line 8: As this is written right now, this is kind of a hack, but it's just a result of how we designed the `Navbar` component

I'm curious what y'alls thoughts are on making this more friendly to page inactivation. We could easily store the `ul` information in a json object and have the Navbar be constructed from it. I think this would also benefit us by making the code DRYer.

pages/events.js
```diff
@@ -0,0 +1,41 @@
...
+				</p>
+
+				<br/><br/>
+				<div className="calendar-container">
+				<iframe src={Config.events.calendar_url}
+					className="calendar-main"
+					frameBorder="0"
+					scrolling="0"
+					title="calendar"
+					/>
```
evanzhong, line 33: Is the choice to remove the calendar embed a design decision, or is there something blocking this on the technical side?

I know we've had discussions about how our current events page is less than ideal, but curious what the forcing function behind this is.

pages/index.js
```diff
@@ -0,0 +1,63 @@
...
+				<div className="content-section">
+					<h2>The largest Computer Science community at UCLA</h2>
+					<p className="subheading">ACM at UCLA is the largest computer science student organization in Southern California. We welcome students of all backgrounds, interests, and skill levels to join our community and share our love for technology. ACM is comprised of eight committees — each serving a different topic and mission. Our events are open to everyone, regardless of major or experience!</p>
+					<SocialMedia />
+				</div>
+				<br />
+				<Committees committees={committees} />
+				<div className="button-section">
+					{ /** CHANGE THIS LINK TO ABOUT PAGE LATER */ }
+					<Link href="/committees"><a className="button">Learn More About Our Committees</a></Link>
```
evanzhong, line 29: What does this comment mean?

styles/components/CommitteeSpread.scss
```diff
@@ -59,6 +59,11 @@
       }
     }
 
+    // .committee-image {
+    //   height: 50px;
+    //   position: relative;
+    // }
+
```
evanzhong, line 66: If you don't foresee this being necessary in the future, you can delete it

styles/pages/committees.scss
```diff
@@ -50,6 +49,23 @@
...
+  // the div is to change display from inline-block to block so there's no weird spacing between border and image
+  // the img is to overwrite the default image shadow from committees.scss to prevent double borders
+  // img should be removable once fully converting to /next/image - the one in committees.scss needs to stay for the non next imgs for now
+  div {
+    display: block !important;
+  }
+
+  img {
+    box-shadow: none !important;
+  }
```
evanzhong, line 66: Interesting stuff, perhaps this is what Bryan mentioned about `next/image` being too abstracted for his liking.

In the future, if this persists we may want to consider creating our own wrapper component for `next/image`

evanzhong (changes requested):
Amazing work @advaithg and @mattxwang!

I mainly looked at the diffs in this review. Most of the relevant items are left in comments in specific files. General things I noticed
- +1 to the comment in the PR writeup, there's a lot of low hanging fruit with using `.modules`; a discussion we should have later after this PR ships.
- There's some extraneous commented out JSX left in files under `TODOs` comments. My personal preference is not to have commented out code in production. If we can always refer back to it in the commit history if needed down the line. I left some comments about this, but didn't comment on every instance I saw.
- We use both regular html `` tags and `next/image`. This could get confusing in future development as devs might not know which is appropriate. I know `next/image` was a particular pain point so we can talk about addressing this in a future issue.

Only thing blocking for me is removal of commented out code, other than that it looks good to go!

pages/index.js
mattxwang: Good catch, I think I added it before refactoring how `.button` worked.

pages/events.js
mattxwang: I believe this is actually from a previous PR I made, #173 - let me know what you think!

(forcing function - the calendar is completely empty)

components/Navbar.js
mattxwang: Yeah, I'm not the largest fan of this either, especially since this requires people to update comments (uncommon) when implementation details change, like they did here.

The JSON implementation is a good idea; our friends at Creative Labs do something similar, see https://github.com/UCLA-Creative-Labs/sunshine/blob/master/components/Navbar.tsx#L9

styles/pages/committees.scss
mattxwang: Yup yup, honestly strange that they force so many style tag changes (causes some specificity shenanigans). Open to any suggestions on how to approach this.

pages/events.js
evanzhong: sgtm, we'll probably axe it anyways when we get to #115

components/Committees/CommitteeEventCard.js
```diff
@@ -1,13 +1,16 @@
+import Image from 'next/image';
 import React from 'react';
 
 function CommitteeEventCard(props) {
     const hasImage = props.image.src;
     return (
         <div className="info-card-container">
+            <div className="image-wrapper">
```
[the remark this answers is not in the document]
mattxwang: I believe so, I think this is from Advaith's side!

components/Carousel.js
```diff
@@ -22,9 +23,7 @@ export default class Carousel extends React.Component {
 						href={item} target="_blank"
 						rel="noreferrer noopener"
 						key={i}>
-							<div
-							style={{backgroundImage: 'url('+ process.env.PUBLIC_URL + item + ')'}}>
-							</div>
+							<Image src={item} width={IMAGE_WIDTH} height={IMAGE_WIDTH}/>
```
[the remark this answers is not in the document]
mattxwang: Yeah, so my reasoning here is that the background image approach itself is a hack - each `div` really semantically should be an image, and without this change (in general) the Carousel would be inaccessible.

I did merge that above change with *also* transitioning it to `<Image />`, which may have jumped the gun. Didn't see any issues with this on prod, but always down to switch back!

mattxwang:
Addressed most of Evan's comments, and updated the PR description with the relevant to-dos!

components/Carousel.js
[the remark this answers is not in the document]
evanzhong: Makes sense, re: accessibility.

If no problems seen on prod with `next/image` I think it's actually a great change. The carousel is a prime place for lazy loading since it's located below the fold and is image dense

evanzhong (approved):
lgtm for this set of changes!

PR #192, opened 2021-06-29
Updates ESLint to ignore the `build` folder

advaithg:
super short PR - the eslint config did not ignore the build files, showing tons of irrelevant errors when running `npm run lint-js` after doing `npm start` at least once, since it was also scanning the build folder

this adds all the js files in the build folder to the eslint, fixing this issue

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,7 @@ module.exports = {
     },
     sourceType: 'module',
   },
-  ignorePatterns: ['/build/**/**/*.js'],
+  ignorePatterns: ['/build/*'],
   plugins: ['react', 'import'],
   settings: {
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     },
     sourceType: 'module',
   },
+  ignorePatterns: ['/build/**/**/*.js'],
```
mattxwang, line 18: Is the entire glob here necessary? Can we just ignore the `build` folder?

advaithg: That would have been easier yes, lemme fix that

mattxwang (approved):

Issue #187, opened 2021-06-29
document CI/CD and Next.js quirks


advaithg:
The current README is incomplete, with a TODO in the place of "Production Deploy"

To fix this, write documentation for the new CI/CD pipeline, and also describe some of the weird quirks with next.js for future devs (mainly `next/image`)

PR #199, opened 2021-07-07
chore: ignores README-only changes for actions triggers

mattxwang:
While working on #187, I noticed that we "waste" a bit of Actions time rebuilding, linting, and running CodeQL against PRs that only introduce changes to `README.md`. Since we aren't using a markdown linter like `remark`, these Actions are all going to be a no-op.

So, this PR disables these actions on commits that only change `README.md`. If this is useful, we may want to:

* expand the set of files that we don't run changes against
* push this change upstream to other react projects we use, particularly the react starters that Teach LA uses

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -14,9 +14,13 @@ name: "CodeQL"
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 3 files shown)


advaithg (approved):
lgtm!

PR #201, opened 2021-07-07
converts about page CSS to use CSS Modules, deals with linting issues

advaithg:
quick change to add css modules to stylelint config
might have to update rules during `.module` conversion as well

edit: also converts `about` page to use `.module` scss

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,8 +1,7 @@
 {
     "extends": [ 
       "stylelint-config-recommended", 
-      "stylelint-config-sass-guidelines", 
-      "stylelint-config-css-modules" 
+      "stylelint-config-sass-guidelines"
     ],
     "plugins": [
       "stylelint-scss"
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,5 +1,9 @@
 {
-    "extends": [ "stylelint-config-recommended", "stylelint-config-sass-guidelines", "stylelint-config-css-modules" ],
+    "extends": [ 
+      "stylelint-config-recommended", 
+      "stylelint-config-sass-guidelines", 
+      "stylelint-config-css-modules" 
+    ],
     "plugins": [
       "stylelint-scss"
     ],
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -17,6 +17,8 @@
       "max-nesting-depth": null,
       "selector-max-compound-selectors": null,
 
-      "color-hex-case": "upper"
+      "color-hex-case": "upper",
+
+      "scss/at-import-partial-extension-blacklist":"never"
     }
   }
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,5 +1,5 @@
 {
-    "extends": [ "stylelint-config-recommended", "stylelint-config-sass-guidelines" ],
+    "extends": [ "stylelint-config-recommended", "stylelint-config-sass-guidelines", "stylelint-config-css-modules" ],
     "plugins": [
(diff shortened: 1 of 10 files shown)


.stylelintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-    "extends": [ "stylelint-config-recommended", "stylelint-config-sass-guidelines" ],
+    "extends": [ "stylelint-config-recommended", "stylelint-config-sass-guidelines", "stylelint-config-css-modules" ],
```
mattxwang, line 2: Any chance we can make this multi-line and get cleaner diffs? May be something we'll have to push down the line, and use a `.js` file so we can have trailing commas.

mattxwang (commented):
Seems good to me; @advaithg, do you have a trivial test case with this config on new module-based code? Would be good to have confidence that we haven't broken CI yet.

mattxwang:
Thanks for the quick change @advaithg, just wanted to check in on this:

> do you have a trivial test case with this config on new module-based code?

Does this work with the about page?

advaithg:
> Does this work with the about page?

Will test it out by merging the about page branch into this one!

mattxwang:
Just to check - now that we're not using the camelCase notation, do we still need `stylelint-config-css-modules`?

advaithg:
> Just to check - now that we're not using the camelCase notation, do we still need `stylelint-config-css-modules`?

doesn't seem like it - lemme get rid of it

mattxwang:
Also, not sure if this is just a GitHub quirk, but the file diff has a few changes that shouldn't be rooted in your commits - have you tried merging `main` into your branch, and/or doing some ~ rebasing ~?

advaithg:
> Also, not sure if this is just a GitHub quirk, but the file diff has a few changes that shouldn't be rooted in your commits - have you tried merging `main` into your branch, and/or doing some ~ rebasing ~?

I think this is from merging the about branch into this branch, which are basically the results of merging main into this. I'll merge main once again just to be safe!

.stylelintrc.json
```diff
@@ -13,6 +16,8 @@
       "max-nesting-depth": null,
       "selector-max-compound-selectors": null,
 
-      "color-hex-case": "upper"
+      "color-hex-case": "upper",
+
+      "scss/at-import-partial-extension-blacklist":"never"
```
mattxwang, line 21: Let's include a comment as to why we made this change, for the future devs!

styles/App.scss
```diff
@@ -25,7 +25,7 @@
 
 /* pages */
 @import 'pages/index';
-@import 'pages/about';
+// @import 'pages/about';
```
mattxwang, line 28: Let's remove this before merging!

mattxwang (commented):
Left some small comments; also, is the fragmentation on the `about` page from an editor setting? Might be good to revert that and let editor soft-wrapping do its work instead.

advaithg:
> is the fragmentation on the `about` page from an editor setting? Might be good to revert that and let editor soft-wrapping do its work instead.

I fragmented it because of the linter, which was giving me issues with lines being too long!

.stylelintrc.json
advaithg: I don't think json files support commenting

mattxwang: Ah great point, thought we had switched it to the `.js` file already. Let's table that 'till later, then.

mattxwang (approved):
Adjusted the PR title, but otherwise seems good to merge. I imagine we'll do another CSS passover later. Great work Advaith!

Issue #188, opened 2021-06-29
standardize images to `next/image`


advaithg:
Currently, a lot of images use the `img` tag, which shouldn't be done in next.js
This issue is to deal with these images and switch to `next/image`

as of now, `img` works as long as `/public` is maintained properly

While it sounds pretty simple, it's a lot more complicated than it sounds because of a bunch of stuff:

-  `next/image` breaks on deploy when the image is an svg file (not sure why)

- netlify doesn't seem to be fully working on deploys with `next/image` (see [this](https://github.com/vercel/next.js/issues/18356))

- `next/image` requires a `height` and a `width` or `layout=fill`, which makes it difficult in certain cases where we want to use the default image size

- `next/image` has some built in CSS that is difficult to override (see `committees.scss` for an example)

This can be put on hold until `next-on-netlify` is updated, as mentioned [here](https://answers.netlify.com/t/new-image-tag-not-working-on-next-js/25571)

mattxwang:
To resolve the manual `img` tags, feel free to search for `<img />`, or 

```js
{/* eslint-disable-next-line @next/next/no-img-element */}
```

PR #216, opened 2021-07-22
Update `committees.js` for NSO Tabling

### LINKED ISSUE #180: Clean up `committees.js`
Now that it's summer, I think we should do something to make `committees.js` less overwhelming (in anticipation of NSO tabling).

In particular, some things to do:

* do we want to keep all the spring events here? if so, should we make it clear?
* if we don't want to keep the spring events, what do we put here instead? 
* images should have `alt` tags! no exceptions, unless you label it as a decorative image!

advaithg:
Resolves #180 

Replaces "Events and Projects" with "Past Events and Projects"
Adds alt tag to every image that doesn't have one, and a general alt for events with no images that use the base committee motif.

### DIFF
--- a/components/Committees/CommitteeEventCard.js
+++ b/components/Committees/CommitteeEventCard.js
@@ -8,7 +8,7 @@ function CommitteeEventCard(props) {
             <div className="image-wrapper">
             {hasImage
                 ? <Image src={props.image.src} alt={props.image.alt} width={241.8} height={160}/>
-                : <Image src={`/images/committees/${props.committee}/${props.committee}_motif_base.png`} alt={props.image.alt} width={241.8} height={160} />
+                : <Image src={`/images/committees/${props.committee}/${props.committee}_motif_base.png`} alt={`base motif for ${props.name}`} width={241.8} height={160} />
             }
             </div>
             <h4 className="font-header text-bold">{props.title}</h4>
--- a/components/Committees/CommitteeSection.js
+++ b/components/Committees/CommitteeSection.js
@@ -13,6 +13,7 @@ function CommitteeProjects({committee}) {
 					const { image, title, desc } = card;
 					return (
 						<CommitteeEventCard
+							name={committee.name}
 							committee={committee.class}
 							image={image}
 							title={title}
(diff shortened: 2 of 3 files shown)


mattxwang:
What are our thoughts on disabling the events section altogether; I'm concerned that this is still information overload.

annaguo1012:
I feel like there is too much information on it right now, but just having the general committee descriptions would be a little vague. What if we kept 3 events/projects for each subcommittee? I feel like they really help give people a better sense of what each committee does and how they could get involved.

annaguo1012 (commented):
I know we might do some more changes, but mostly everything looks good to me! There are a few warnings though to use Image from 'next/image' instead of . Do we want to go with that or keep using ? It says Next.js' image component has better performance and automatic image optimization. It also still does lazy-loading by default.

advaithg:
> What are our thoughts on disabling the events section altogether; I'm concerned that this is still information overload.

@mattxwang I think we can go with @annaguo1012 idea, and keep 3 events/projects per committee instead of keeping all of them. I'm kind of concerned about completely disabling events, since that means there's no tangible examples of what each committee does.

advaithg:
> I know we might do some more changes, but mostly everything looks good to me! There are a few warnings though to use Image from 'next/image' instead of . Do we want to go with that or keep using ? It says Next.js' image component has better performance and automatic image optimization. It also still does lazy-loading by default.

The reason we use `` instead of `/next/image` is because `/next/image` has issues when deploying on netlify and when it's applied on SVGs, so throughout the entire codebase - all the PNGs use `/next/image`, but we're stuck with `img` for the SVGs, which include the committee logos and some other images.

mattxwang:
Sounds good, I'll let you and Anna work on this then - great idea @annaguo1012!

mattxwang:
> There are a few warnings though to use Image from 'next/image' instead of 

If this makes code review hard, we can disable the rule just for the lines that trigger the warning, and TODO removing them + link them to #188!

annaguo1012 (approved):
LGTM!

mattxwang:
Just wanted to check, is this the correct image for JS Chats?

Screen Shot 2021-07-25 at 3 51 37 PM

annaguo1012:
Oh no, that's not the right one. Sorry for not catching that! I can fix it now.

Issue #112, opened 2021-04-18
Intern(ship) program page


mattxwang:
One of the more common pieces of feedback we got was that the intern program was confusing to navigate! I think the website is one of the best ways we can clear up this confusion; we should make a page that outlines everything any interested party should know.

Working on this ticket is much like working on a bigger informational page for a large company: you'll work with various parties to balance everyone's needs. The general workflow for this ticket should look something like:

1. first, meet with Ava Asmani, the new internship director; she has a certain vision for this page
2. then, create a barebones skeleton of the relevant layout/functionality
3. with draft copy, flesh out individual sections component by component - placeholder content is fine
4. once the page is laid out, work with ACM Design & Ava on the UX: is it intuitive? easy to navgiate? hierarchical content?
5. if possible, a bit of 'user testing' - asking previous interns, current officers, etc.
6. finalize content, copy
7. publish 🚀 

While this is a quite involved process, you (the dev) are mostly responsible for the layout of the page and the mapping of the content, but not writing the content yourself. If you are interested in copywriting or web design, this is a great way to dip your toes into the work process.

Some general content that this site should have:

* what *is* the internship program? how does it differ from being a general member, or an officer? does it matter per-committee? dev team?
    * what information is most important? How should the page layout reflect this information hierarchy? This page will serve as the internship program's external facing image, how do we want to represent ourselves?
* when is the timeline for application? for interviews? for acceptance? (this is obv TBD for next year)
* who is eligible? do you need to have prior experience for any internships?
* do we pay people? (no)
* other relevant FAQs (Ava is likely the best point person for this)

Some goodies you can debate on:

* in general, how dynamic or static should this page be?
* do we want to feature past interns? if so, how - with photos, testimonials, etc.?
* how much information should be on this page, opposed to committee websites; what if committees don't have a website?
* how to we encode the data that goes *into* this page? keep in mind that often times, people who don't have familiarity with our tech stack may want to edit this page.
* how do we ensure that this page especially is accessible to everyone? (font readability, color contrast, etc).

mattxwang:
Just to update - met with Ava, and she wants to work on this starting in the summer! @advaithg will be the lead dev, and we'll try to pull someone from Design to help us with the UX.

mattxwang:
Closed by #222.

Issue #191, opened 2021-06-29
CSS cleanup and `.module` SASS


advaithg:
We have a lot of repetitive CSS, some of which can be cleaned up or made more generalized.

Using `.module` SASS should deal with some of this repetition, and works better with next.js as well

Leaving it to @mattxwang to elaborate!

PR #222, opened 2021-07-26
Adds Internship Page!

advaithg:
Addresses issue #112 

Preliminary Design [here](https://www.figma.com/file/RQT8mBsi0U13W8cJsymV5b/Internship-Page-Mockup)!

Aims to explain how ACM's internship program works to prospective interns, giving details about each committee's internship program, the expectations, and the next steps.

Adds `internship.js` in `pages` as well as in `data`, `internship.scss` and a number of component js and scss files (`NextSteps`, `CommitteeInternshipInfo`, `InternshipTimeline`).
Also adds card assets for all the committees and board under `public/images/cards` from the Figma.

Pending adding data, testimonials and FAQ along with feedback from previous and prospective interns and Ava, the internship director!


components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,67 @@
+import Image from 'next/image';
+import React, {useState} from 'react';
+
+function CommitteeInternshipInfo(props) {
+    const committees = props.committees;
+    const [committeeName, setCommitteeName] = useState('Board');
+    return(
+        <div>
```
mattxwang, line 8: Is there a reason why there's this parent div? Worried about having too many divs haha, esp wrt accessibility tree.

components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,67 @@
...
+            {committees.map(committee =>
+                committeeName == committee.name ?
+                <div className={'committee-internship-info ' + committee.class}key={committee.name}>
+                    <Image
+                        src={committee.image}
+                        alt={`${committee.name}'s wordmark`}
+                        height="60px"
+                        width="250px"
+                    />
+                    {/* <img src={committee.image} alt="committee logo"/> */}
```
mattxwang, line 39: Can we remove this before merging?

components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,67 @@
+import Image from 'next/image';
+import React, {useState} from 'react';
+
+function CommitteeInternshipInfo(props) {
```
mattxwang, line 4: One thought on this function - if we'd like, we can create a `board` committee entry (i.e. an object with the relevant keys), and then just prepend/append it to the props; this can let us reduce duplication, since the implementation itself can rely on just interpreting `committees` instead of creating a special case for board.

components/Internship/InternshipTimeline.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+function InternshipTimeline(props) {
+	const { items } = props;
+	const totalItems = items.length;
+	const numberOfActiveItems = items.filter(item => item.active).length;
+	const progressBarWidth = totalItems > 1 ? (numberOfActiveItems - 1) / (totalItems - 1) * 100 : 0;
+
+	return (
```
mattxwang, line 9: Just checking, are all of these divs necessary? And if so, can we use some sort of semantic HTML, aria labelling, or something along those lines to make this more accessible?

For example, the progress bar - at the very least, let's annotate this somehow so users with screenreaders can still "see" the progress of the internship timeline.

components/Internship/NextSteps.js
```diff
@@ -0,0 +1,25 @@
+import Image from 'next/image';
+import React, {useState} from 'react';
+
+function NextSteps(props) {
+    const { image, name, info } = props;
+    const [isFlipped,setFlipped] = useState(true);
```
mattxwang, line 6: Did eslint catch this comma space? i.e.

```js
const [isFlipped, setFlipped]
```

components/Internship/NextSteps.js
```diff
@@ -0,0 +1,25 @@
...
+    return (
+        <div className="committee-card" onClick={() => setFlipped(!isFlipped)}>
+            <div className={isFlipped ? 'next-steps-card-inner' : 'next-steps-card-inner is-flipped'}>
+                <div className='next-steps-card-face'>
+                    <Image
+                        src={image}
+                        alt={`${name}'s card motif`}
```
mattxwang, line 13: motif is not a great alt text tag, we might want to workshop this!

components/Internship/NextSteps.js
```diff
@@ -0,0 +1,25 @@
+import Image from 'next/image';
+import React, {useState} from 'react';
+
+function NextSteps(props) {
+    const { image, name, info } = props;
+    const [isFlipped,setFlipped] = useState(true);
+    return (
+        <div className="committee-card" onClick={() => setFlipped(!isFlipped)}>
```
mattxwang, line 8: Since this is clickable, we'll want to somehow mark this as a clickable/tabbable element - either a `<button>` (which imo is maybe inappropriate here) or some sort of `role` and `aria` labels. Would encourage you to check out how some of the bigger players implement this!

components/Internship/NextSteps.js
```diff
@@ -0,0 +1,25 @@
...
+            <div className={isFlipped ? 'next-steps-card-inner' : 'next-steps-card-inner is-flipped'}>
+                <div className='next-steps-card-face'>
+                    <Image
+                        src={image}
+                        alt={`${name}'s card motif`}
+                        layout='fill'
+                    />
+                </div>
+                <div className='next-steps-card-face back'>
```
mattxwang, line 17: This should probably semantically be a `<p>` tag? But unsure, this is a judgement call on your end.

data/committees.js
```diff
@@ -57,6 +58,19 @@ export default [
 		],
 		introImage: '/images/committees/studio/studio_motif_applied.png',
 		mission: 'Our mission is to teach skills revolving around video game development such as computer science, game design and art in order to help usher students into the game development industry.',
+		whatInternsWillDo:
```
mattxwang, line 61: This comment applies for all the committees stuff - are you looking to have bullet points, etc.? If so, maybe this should be a JSX element.

data/committees.js
```diff
@@ -57,6 +58,19 @@ export default [
...
+		hosting tutorials in the winter quarter (which we haven't done in the past). Interns will not be expected to lead tutorials, but they will be expected to familiarize themselves with the tutorial content by attending the 
+		respective subcommittee meetings, making sure to ask the tutorial host questions before the event to ensure that the intern can help run the tutorial during the event. Prior skills in the Unity engine (which is Studio's
+		primary form of game programming) is heavily valued for this role but not entirely needed, as much of game development is a "learn as you go" type of style. Strong skills in C# (very similar to C++) and teaching 
+		skills are also appreciated, but these are all soft expectations.`,
+		timeCommitment: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Lectus nulla at volutpat diam ut. Elementum nisi quis eleifend quam adipiscing vitae.',
+		nextSteps: `An applicant fills out an officer interest form with various questions regarding skillset/reason for applying, then we schedule a 15-minute interview
```
mattxwang, line 71: These keys may be less obvious for someone editing - do we want to nest this under a `internship` key?

data/internship.js
```diff
@@ -0,0 +1,123 @@
+export default {
+    items: [
+        {
+            date: '00/00',
+            name: 'apps open',
+            info: 'information about the event',
+            active: true,
```
mattxwang, line 7: If we have the date, could we set this programmatically?

data/internship.js
```diff
@@ -0,0 +1,123 @@
...
+        },
+    ],
+    QA: {
+        rows: [
+            {
+                title: 'Lorem ipsum dolor sit amet,',
+                content: `Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sed tempor sem. Aenean vel turpis feugiat,
+                ultricies metus at, consequat velit. Curabitur est nibh, varius in tellus nec, mattis pulvinar metus.
+                In maximus cursus lorem, nec laoreet velit eleifend vel. Ut aliquet mauris tortor, sed egestas libero interdum vitae.
+                Fusce sed commodo purus, at tempus turpis.`,
```
mattxwang, line 85: Similar comment - should this be JSX?

data/internship.js
```diff
@@ -0,0 +1,123 @@
...
+        breakpoint: { max: 1024, min: 464 },
+        items: 2,
+        slidesToSlide: 2, // optional, default to 1.
+        },
+        mobile: {
+        breakpoint: { max: 464, min: 0 },
+        items: 1,
+        slidesToSlide: 1, // optional, default to 1.
+        },
+    },
```
mattxwang, line 122: It seems like we have a few magic numbers / colors here - can we document this and/or add some `const` vars to understand what's going on here?

Also, indentation seems a bit off?

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                <h1>Internships</h1>
+                What do you do as an intern? Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+                sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Tristique sollicitudin
+                nibh sit amet commodo nulla. Amet facilisis magna etiam tempor.
+            </div>
+            <CommitteeInternshipInfo
+                committees={committees}
+            />
+            <div className="content-section text-center">
+                <h1>Applications</h1>
```
mattxwang, line 32: This should be an `<h2>`, there's an `<h1>` above this already.

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                <InternshipTimeline items={items}/>
+                <div className="application-information">
+                    <div>
```
mattxwang, line 35: What is this `<div>` used for?

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                        {items.map((item, i) => (
+                        <div key={i} className='application-item'>
```
mattxwang, line 37: Similarly, should this semantically be another tag?

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                                <a className="button" target="_blank" rel="noreferrer noopener">
+                                    Apply Now
+                                </a>
+                            </Link>
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <div className="content-section text-center">
+                <h1>Next steps</h1>
```
mattxwang, line 58: Similar comment on heading specificity, should not be an `<h1>`

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                            {item.name}
+                            <p>{item.info}</p>
+                        </div>
+                        ))}
+                        -- more dates will be added throughout the quarter --
+                    </div>
+                    <div className='application-info-card'>
+                        <h3>Internship Applications Reopen Fall 2021</h3>
+                        <div className='application-button'>
+                            <Link href="">
```
mattxwang, line 48: We shouldn't have blank hrefs; at the very least, should mark this with a TODO!

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                            image={committee.card}
+                            name={committee.name}
+                            key={committee.name}
+                            info={committee.nextSteps}
+                        />,
+                    )}
+                </Carousel>
+            </div>
+            <div className="content-section text-center">
+                <h1>Past Intern Testimonials</h1>
```
mattxwang, line 87: Same comment on header specificity.

pages/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                                    height = "60px"
+                                /> */}
+                                <p>{testimonial.quote}</p>
+                                <p className='testimonial-giver'>- {testimonial.name}, {testimonial.committee} intern, {testimonial.year}</p>
+                            </div>
+                    ))}
+                </div>
+            </div>
+            <div className="content-section text-center">
+                <h1>FAQ</h1>
```
mattxwang, line 104: Header specificity!

styles/App.scss
```diff
@@ -22,6 +22,9 @@
 @import 'components/SocialMedia';
 @import 'components/TechGala/ProjectCard';
 @import 'components/TechGala/ViewBanner';
+@import 'components/Internship/CommitteeInternshipInfo';
+@import 'components/Internship/InternshipTimeline';
+@import 'components/Internship/NextSteps';
```
mattxwang, line 27: Can we module these instead? For new CSS, unless there's a good reason let's module the files so we don't have to deal with it *again* in #191.

styles/App.scss
```diff
@@ -30,3 +33,4 @@
 @import 'pages/events';
 @import 'pages/sponsors';
 @import 'pages/techgala';
+@import 'pages/internship';
```
mattxwang, line 36: Same comment on modules; if the card CSS is reusable, might be good to move that into a component file.

styles/components/Internship/CommitteeInternshipInfo.scss
```diff
@@ -0,0 +1,79 @@
...
+  border-radius: 1rem;
+  box-shadow: 0 0.125rem 0.5rem rgba(0, 0, 0, 0.3), 0 0.0625rem 0.125rem rgba(0, 0, 0, 0.2);
+  font-size: 1.25em;
+  margin: auto;
+  margin-bottom: 50px;
+  padding: 1.125em 1.5em;
+  position: relative;
+  width: 80vw;
+  
+  &.studio {
```
mattxwang, line 12: Can we do this programatically with a SASS loop / mixin instead of duplicating the code like this?

styles/components/Internship/CommitteeInternshipInfo.scss
```diff
@@ -0,0 +1,79 @@
...
+  margin: auto;
+  margin-top: 50px;
+  padding-bottom: 10px;
+  text-align: center;
+  width: 80vw;
+  
+  .logo-diamond {
+    padding: 5px;
+  
+    img {
```
mattxwang, line 62: Can we avoid element selectors if possible? This is nested, so not as big of a deal, but would prefer explicit markup.

styles/components/Internship/CommitteeInternshipInfo.scss
```diff
@@ -0,0 +1,79 @@
...
+      &:hover {
+        cursor: pointer;
+        transform: scale(1.08);
+      }
+    }
+  }
+}  
+  
+@media screen and (max-width: 850px) {
```
mattxwang, line 71: Do we want to use an official breakpoint (i.e. a variable) here, or is there a reason why we're using `850px`?

styles/components/Internship/NextSteps.scss
```diff
@@ -0,0 +1,51 @@
...
+    width: 100%;
+  }
+  
+  &.is-flipped {
+    transform: rotateY(180deg);
+  
+    img {
+      visibility: hidden;
+    }
+  }
```
mattxwang, line 26: Same comment on element selectors!

styles/components/Internship/InternshipTimeline.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+  
+.timeline-item {
+  position: relative;
+}
+  
+.timeline-item::before {
+  background-color: #CCC;
+  border-radius: 100%;
+  content: '';
```
mattxwang, line 29: Could we leave some comments on what this code is doing? Stuff with pseudoselectors can definitely be confusing!

styles/components/Internship/InternshipTimeline.scss
```diff
@@ -0,0 +1,77 @@
...
+.timeline-item.active .timeline-content {
+  background-color: $acm-cobalt;
+  color: #FFF;
+}
+  
+.timeline-item.active .timeline-content::before {
+  background-color: $acm-cobalt;
+}
+
+@media screen and (max-width: 700px) {
```
mattxwang, line 73: Same comment on breakpoints.

styles/pages/internship.scss
```diff
@@ -0,0 +1,88 @@
...
+
+p {
+  font-size: 0.8em;
+  line-height: 150%;
+  margin: 0 22px 14px;
+
+  &.subheading {
+    font-size: 1em;
+  }
+}
```
mattxwang, line 15: Can we avoid using element selectors - this is particularly pernicious since it's globally-scoped, and now import order will break specificity ties.

styles/pages/internship.scss
```diff
@@ -0,0 +1,88 @@
...
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+}
+
+.application-item {
+  margin-bottom: 15px;
+  margin-top: 10px;
+  text-align: left;
+
+  p {
```
mattxwang, line 27: Same comment on element selectors.

styles/pages/internship.scss
```diff
@@ -0,0 +1,88 @@
...
+
+  &.subheading {
+    font-size: 1em;
+  }
+}
+
+.application-information {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+}
```
mattxwang, line 20: Does a similar class to this already exist in globals?

styles/pages/internship.scss
```diff
@@ -0,0 +1,88 @@
...
+
+.testimonial-section {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr;
+  margin-top: 20px;
+}
+
+.testimonial-giver {
+  text-align: right;
+}
```
mattxwang, line 58: I believe we have a `text-right` class for this now, or coming soon?

styles/pages/internship.scss
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+@media screen and (max-width: 800px) {
+  .testimonial-section {
+    grid-template-columns: 1fr 1fr;
+  }
+}
+
+@media screen and (max-width: 500px) {
```
mattxwang, line 74: Same comments on breakpoints

styles/pages/internship.scss
```diff
@@ -0,0 +1,88 @@
...
+}
+
+.testimonial-giver {
+  text-align: right;
+}
+
+.faq {
+  .faq-row-wrapper {
+    .faq-body {
+      text-align: left;
```
mattxwang, line 63: Could we just use `.text-left` or something similar here?

mattxwang (changes requested):
Whew, this is a long one! Here's a first pass, I think this'll be a long-ish back-and-forth.

Let me know if you have any q's - had to be a bit brief just since I'm juggling a ton of stuff right now.

components/Internship/CommitteeInternshipInfo.js
advaithg: there's an error without this since not everything is within a single div! the div right below this is for the line of logos.

components/Internship/CommitteeInternshipInfo.js
advaithg: Hm I tried doing this, but there seems to be a problem with `useEffect` in that it doesn't run automatically when the page loads but only once I click on one of the other logos - will push these changes in a bit!

advaithg: (since it isn't running on page load, it isn't showing up in the next steps cards either)

data/internship.js
advaithg: hm in this case the numbers I went with are the recommended values for the carousel package I imported - not sure how to document that

pages/internship.js
advaithg: `application-information` is a grid, so this is one of the columns!

styles/pages/internship.scss
advaithg: can't! it's a component from a package, so I have to use their class hierarchy to edit any styles they don't let us edit directly using `FAQStyles`

components/Internship/CommitteeInternshipInfo.js
mattxwang: Why would you need to use `useEffect` here? I mean running this command on build, i.e. creating a new object at the top-level and prepending.

components/Internship/CommitteeInternshipInfo.js
mattxwang: In that case, you should use a [Fragment](https://reactjs.org/docs/fragments.html) - the empty tag `<> </>` should be fine for this use-case.

styles/pages/internship.scss
mattxwang: Oh I see. In that case, let's add a comment that explains this, and links to the relevant component - since that component might change in the future, it would be good to know what we'd need to update.

components/Internship/CommitteeInternshipInfo.js
advaithg: directly doing that for some reason duplicates the `board` entry, adding it twice (and sometimes more times) - I'm not sure why it's happening but I think it's something to do with it prepending again on each render - which is why I switched to `useEffect`?

components/Internship/NextSteps.js
advaithg: hmm i wasn't really sure what to name it but yeah motif probably isn't a great alt. Not sure how to come up with a generalized alt for this other than using 'motif' or 'image' though. Maybe I can write descriptions for each card and add an `altTextCard` property to the js data file to be used as the alt tag?

components/Internship/CommitteeInternshipInfo.js
mattxwang: So, you'd run this top-level (i.e. not in the render function of the component) - that way, the expression gets run once when the variable is first evaluated, and that would be it. That way, you shouldn't need `useEffect`, and the expression gets run and built at build time.

data/internship.js
mattxwang: I would then mention exactly that, and link to the docs!

pages/internship.js
mattxwang: Gotcha. Might be good to try to incorporate a `<section>` somehow in this hierarchy, there's a ton of nested divs! That being said, let's say that this isn't a blocking concern for now.

styles/App.scss
advaithg: I can module the entire page, but I'll have to mess with `index.scss` and `global.scss`, since i use `content-section` and it's in index - whereas it should be in globals
If that won't interfere with any other PRs working on this, I'll go ahead and do it!
Another option is to for now duplicate the `content-section` code I need into `Internship.module.scss` and then module it - but that feels like unnecessary extra steps.

styles/pages/internship.scss
advaithg: don't have one, but I added one in globals!

styles/components/Internship/CommitteeInternshipInfo.scss
advaithg: the reason i didn't use a SASS loop is because the color is labelled differently for each committee. Not sure how I'd use mixins either because of the unique color name for each committee.

pages/internship.js
```diff
@@ -0,0 +1,147 @@
...
+import CommitteeInternshipInfo from '../components/Internship/CommitteeInternshipInfo';
+import InternshipTimeline from '../components/Internship/InternshipTimeline';
+import NextSteps from '../components/Internship/NextSteps';
+import Layout from '../components/Layout';
+
+import data from '../data';
+import 'react-multi-carousel/lib/styles.css';
+
+const { committees, internship } = data;
+committees.unshift(
```
mattxwang, line 17: Quick note - try not to use `committees.unshift`, since you're modifying the array in-place (and I think causing some other shenanigans - try going to the internship page, and then switching to the committees page. I get an error).

Instead, I suggest concatenating it, which may also involve deep-copying:

```js
const boardInfo = {...};
const internshipInfo = boardInfo.concat(committees);
```

More on `concat` in the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat).

pages/internship.js
advaithg: not sure which tag I'd use here!

components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,42 @@
...
+                            src={committee.diamond}
+                            alt={`${committee.name}'s logo`}
+                            width="100px"
+                            height="100px"
+                        />
+                    </div>,
+                )}
+            </div>
+            {committees.map(committee =>
+                committeeName == committee.name ?
```
mattxwang, line 24:
```suggestion
                committeeName === committee.name ?
```

components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,42 @@
...
+import styles from '../../styles/components/Internship/CommitteeInternshipInfo.module.scss';
+
+function CommitteeInternshipInfo(props) {
+    const {committees} = props;
+    const [committeeName, setCommitteeName] = useState('Board');
+    return(
+        <>
+            <div className={styles['logos-line']}>
+                {committees.map(committee =>
+                    <div className={styles['logo-diamond']} key={committee.name} onClick={()=>setCommitteeName(committee.name)}>
```
mattxwang, line 13: If this is a clickable item, can we make it a `<button>`?

There's more accessibility work that needs to be done here. The clickable elements aren't tabbable, and there's some HTML annotations we need to make; see [this W3 example of a tab implementation](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-2/tabs.html).

components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,42 @@
...
+                        <Image
+                            src={committee.diamond}
+                            alt={`${committee.name}'s logo`}
```
mattxwang, line 16: In this case, this is not good alt text, since we really just want this to read as "ACM Hack", etc.

components/Internship/CommitteeInternshipInfo.js
```diff
@@ -0,0 +1,42 @@
...
+                <div className={`${styles['committee-internship-info']} ${styles[committee.class]}`} key={committee.name}>
+                    <Image
+                        src={committee.image}
+                        alt={`${committee.name}'s wordmark`}
+                        height="60px"
+                        width="250px"
+                    />
+                    <h4>What do we do?</h4>
+                    <p className={styles['internship-info-data']}> {committee.internship.whatWeDo} </p>
+                    <h4>What will interns do?</h4>
```
mattxwang, line 34: Are these supposed to be `<h4>` tags? I don't see an `<h3>` anywhere.

components/Internship/InternshipTimeline.js
mattxwang: Bump on this.

components/Internship/NextSteps.js
```diff
@@ -0,0 +1,27 @@
+import Image from 'next/image';
+import React, {useState} from 'react';
+
+import styles from '../../styles/components/Internship/NextSteps.module.scss';
+
+function NextSteps(props) {
+    const { image, name, info } = props;
+    const [isFlipped, setFlipped] = useState(true);
+    return (
+        <div className={styles['committee-card']} onClick={() => setFlipped(!isFlipped)}>
```
mattxwang, line 10: Same comment on buttons and clickable elements. Currently not tabbable.

components/Internship/NextSteps.js
mattxwang: So the idea is that if the image doesn't render (or the user cannot see the image), then the system should be usable. With that in mind, "motif" is not actually the relevant description: it should be something along the lines of "clicking this will let you see more information".

components/Internship/NextSteps.js
```diff
@@ -0,0 +1,27 @@
...
+            <div className={`${styles['next-steps-card-inner']} ${styles[isFlipped ? null : 'is-flipped']}`}>
+                <div className={styles['next-steps-card-face']}>
+                    <Image
+                        src={image}
+                        alt={`${name}'s card motif`}
+                        layout='fill'
+                    />
+                </div>
+                <div className={`${styles['next-steps-card-face']} ${styles.back}`}>
```
mattxwang, line 19: This should probably be a `p` tag, or at least something else!

data/committees.js
```diff
@@ -57,6 +57,20 @@ export default [
...
+			whatWeDo: 'Studio wants to create a community of game developers, whether you’re interested in gameplay programming, graphics, animation, pixel, music, whatever. We want to help you express yourself through video game development.',
+			whatInternsWillDo: (
+			<ul>
+				<li> Marketing interns will focus on community engagement,
+					particularly in the form of leading socials and increasing member retention</li>
+				<li> Unity Tutorial Subcommittee interns will help host Unity tutorials and bugfix in winter quarter.
+					Unity experience is heavily valued but not absolutely required, as much of game development is a
+					&apos;learn as you go&apos; style. C# skills and teaching experience are appreciated but are soft
+					expectations.</li>
+			</ul> ),
```
mattxwang, line 70: Let's disable the eslint rule enforcing max line length, either in the file entirely or in front of each relevant line. This is going to destroy our ability to preserve meaningful commits.

data/internship.js
```diff
@@ -0,0 +1,146 @@
...
+            image: '.',
+            committee: 'Teach LA',
+            year: '2020-2021',
+        },
+    ],
+    QA: {
+        rows: [
+            {
+                title: 'What is the Internship Program?',
+                content: `The ACM Internship program is a program that facilitates the transition of UCLA students to ACM officers.
```
mattxwang, line 85: Why are content strings backticked if there's no string interpolation?

pages/internship.js
```diff
@@ -0,0 +1,152 @@
...
+import CommitteeInternshipInfo from '../components/Internship/CommitteeInternshipInfo';
+import InternshipTimeline from '../components/Internship/InternshipTimeline';
+import NextSteps from '../components/Internship/NextSteps';
+import Layout from '../components/Layout';
+
+import data from '../data';
+import 'react-multi-carousel/lib/styles.css';
+
+const { committees, internship } = data;
+const boardInfo = [
```
mattxwang, line 17: Can we put this in a file?

pages/internship.js
mattxwang: `<div>` is the most generic tag. [Here's a quick W3Schools article](https://www.w3schools.com/html/html5_semantic_elements.asp) on some options.

pages/internship.js
```diff
@@ -0,0 +1,152 @@
...
+        <Layout>
+            <Banner decorative />
+            <div className="content-section text-center">
+                <h1>Internships</h1>
+                The ACM Internship program is a program that facilitates the transition of UCLA students to
+                ACM officers. Interns learn about the ACM committee they are interning for and ACM in general
+                before becoming an official ACM officer.
+                <br/>
+                Click on the different ACM subcommittee icons in order to learn more about what interns in each
+                committee do and scroll down to find more information on how you can apply to the Internship Program.
```
mattxwang, line 72: This content should be wrapped in a tag, like `<p>`. We shouldn't use `<br />` purely for spacing.

Also, same comment on disabling max line length.

pages/internship.js
mattxwang: Going to unresolve, since we're merging this before the application comes out.

pages/internship.js
```diff
@@ -0,0 +1,152 @@
...
+
+const internshipInfo = boardInfo.concat(committees);
+
+function InternshipPage(){
+    const { items, testimonials, QA, FAQStyles, responsive } = internship;
+    return(
+        <Layout>
+            <Banner decorative />
+            <div className="content-section text-center">
+                <h1>Internships</h1>
```
mattxwang, line 66: How do we feel about 

```suggestion
                <h1>The Internship Program</h1>
```

data/internship.js
```diff
@@ -0,0 +1,146 @@
...
+                Accidentally breaking tables at the intern retreat
+                <br/><br/><b>How much CS experience did you have before you became an intern?</b><br/>
+                Not much - I took AP CS in high school and didn’t learn a ton,
+                and when I applied to be an intern I was taking CS 31.
+                <br/><br/><b>Are you still an officer in ACM? If so, in what Committee?</b><br/>
+                Yes! I’m the director of ACM Impact!
+                <br/><br/><b>What advice would you give to prospective interns?</b><br/>
+                Try to maximize the amount of rejections you get - try everything!
+            </>,
+            image: '.',
```
mattxwang, line 56: What is the point of this key?

data/internship.js
```diff
@@ -0,0 +1,146 @@
...
+            quote: <>
+                <b>What is your favorite memory from being an intern?</b><br/>
+                Accidentally breaking tables at the intern retreat
+                <br/><br/><b>How much CS experience did you have before you became an intern?</b><br/>
+                Not much - I took AP CS in high school and didn’t learn a ton,
+                and when I applied to be an intern I was taking CS 31.
+                <br/><br/><b>Are you still an officer in ACM? If so, in what Committee?</b><br/>
+                Yes! I’m the director of ACM Impact!
+                <br/><br/><b>What advice would you give to prospective interns?</b><br/>
+                Try to maximize the amount of rejections you get - try everything!
```
mattxwang, line 54: Instead of using `<br/>` and `<b>`, let's (at the very least) use `<p>` and header tags.

styles/App.scss
mattxwang: You can use global code without it being moduled.

For example,

```js
<div className={`content-section ${styles['some-module-css-class']}`}>
...
```

styles/components/Internship/CommitteeInternshipInfo.module.scss
```diff
@@ -0,0 +1,85 @@
...
+    box-shadow: 0 0.125rem 0.5rem $ai-arctic, 0 0.0625rem 0.125rem $ai-arctic;
+  }
+  
+  &.hack {
+    box-shadow: 0 0.125rem 0.5rem $hack-purple, 0 0.0625rem 0.125rem $hack-purple;
+  }
+  
+  &.board {
+    box-shadow: 0 0.125rem 0.5rem $acm-cobalt, 0 0.0625rem 0.125rem $acm-cobalt;
+  }
```
mattxwang, line 54: Would like to generate this with a loop if we can.

styles/components/Internship/InternshipTimeline.module.scss
```diff
@@ -0,0 +1,79 @@
+@import '../../variables.scss';
+
+.timeline {
+  background-color: #CCC;
+  height: 4px;
+  margin: 50px 60px;
+  margin-bottom: 125px;
```
mattxwang, line 7: Can we combine this into one `margin`?

styles/components/Internship/InternshipTimeline.module.scss
```diff
@@ -0,0 +1,79 @@
...
+  
+.timeline-item.active .timeline-content::before {
+  background-color: $acm-cobalt;
+}
+
+@media screen and (max-width: $tablet-breakpoint) {
+  .timeline {
+    display: none;
+  }
+}
```
mattxwang, line 79: Can we instead create a generic `.hidden-mobile`?

styles/components/Internship/NextSteps.module.scss
```diff
@@ -0,0 +1,41 @@
+.next-steps-card-inner {
+  background-color: #FFF;
+  border-radius: 1rem;
+  box-shadow: 0 0.125rem 0.5rem rgba(0, 0, 0, 0.3), 0 0.0625rem 0.125rem rgba(0, 0, 0, 0.2);
```
mattxwang, line 4: Stretch goal - mixin this.

styles/variables.scss
```diff
@@ -46,3 +46,4 @@ $font-secondary-color: #777;
 
 $navbar-breakpoint: 1160px;
 $tablet-breakpoint: 768px;
+$single-column-breakpoint: 500px;
```
mattxwang, line 49: What's the point of this variable? If we only use it in a couple of places, we can always just define it scoped in that file (instead of globally).

data/internship.js
```diff
@@ -0,0 +1,146 @@
...
+            {
+                title: 'How can I get involved in ACM if I am not an intern?',
+                content: `If you are not an intern, there are plenty of ways to get involved in ACM. You can always attend workshops that different 
+                committees are hosting. Subscribe to our Newsletter if you would like weekly updates on when workshops are. Some committees, like 
+                Teach LA, allow students to be officers/contributing members without participating in the ACM Internship Program. `,
+            },
+            {
+                title: 'I have more questions, how can I get them answered?',
+                content: `If you have any questions about the application process or the Internship program in general email Ava Asmani, 
+                the Internship Director[ava24@g.ucla.edu]. If you have any questions that are specific to committees, there will be an
```
mattxwang, line 115: This should be a link, right? Or at least formatted differently?

data/internship.js
```diff
@@ -0,0 +1,146 @@
...
+                title: 'How can I get involved in ACM if I am not an intern?',
+                content: `If you are not an intern, there are plenty of ways to get involved in ACM. You can always attend workshops that different 
+                committees are hosting. Subscribe to our Newsletter if you would like weekly updates on when workshops are. Some committees, like 
+                Teach LA, allow students to be officers/contributing members without participating in the ACM Internship Program. `,
+            },
+            {
+                title: 'I have more questions, how can I get them answered?',
+                content: `If you have any questions about the application process or the Internship program in general email Ava Asmani, 
+                the Internship Director[ava24@g.ucla.edu]. If you have any questions that are specific to committees, there will be an 
+                Internship Program ACM Committee Info Session/QandA sometime in week 3 or 4 of the Fall Quarter.`,
```
mattxwang, line 116:
```suggestion
                Internship Program ACM Committee Info Session/Q&A sometime in week 3 or 4 of the Fall Quarter.`,
```

mattxwang (changes requested):
Leaving some relatively strict comments;

First things first, can you merge `main` into this branch? There's quite a few changes, and in particular, I've added an accessibility auditor in the console (via `axe`) - would love to see what that tool says about the new page.

In addition, can we use any of the carousel that Tanaya has implemented for the Impact page? I'm not a huge fan of this library (or in general, adding any libraries when they're not needed).

Some visual/content things:

* the committee logo + wordmark in the "what do we do, what do interns do" section is not the same (height) size. can we adjust that?
* how much can we link to things in the "what do we do, what do interns do" section? there's a lot of verbiage that new people wouldn't be familiar with
* I'm assuming we're rolling back the timeline (since none of these events have happened yet) 
* What does the Apply Now button do right now?
* what is going in to "-- more dates will be added throughout the quarter --"?
* On "Intern to Officer Transition" - it's not obvious at all that I have to click on the card to flip it. Plus, if you're not familiar with the logos, then you may not know what each committee is. How does this behave on mobile? Can we make the border a bit lighter?
* In "Hear from past interns!", the spacing & styling is a bit off. I think we need more padding at the top; the text color and weight should change for headers, and we may want to put the name / role / year fixed at the bottom of each card.

See this for Cyber's card:

Screen Shot 2021-09-15 at 9 36 19 PM

pages/internship.js
advaithg: it does sound better yep, going to change it!

styles/components/Internship/CommitteeInternshipInfo.module.scss
advaithg: already said this - but im not sure how to do this since the loop isn't constant due to the color names being different for each committee

styles/variables.scss
advaithg: hm okay I'll move it to the specific file then

advaithg: it's in here now since I thought it would come in useful later, but I guess we could always move it back to variables if we needed to

data/internship.js
advaithg: I don't want the `<p>` formatting, so I think I'm going to use just headers instead
or I could do `<p>` and add a class to modify the formatting if you think there should be a `<p>`

components/Internship/NextSteps.js
advaithg: I think the common way is to say `role="button"` to mark it as clickable - so I did that. Will look into tab roles when I add tabbing!

data/internship.js
mattxwang: If you want different formatting, I would just override the style - tags are fundamentally about the semantic meaning of the content, not the styling itself.

components/Internship/NextSteps.js
advaithg: this shouldn't be a `<p>`, since it's for the entire card face - not just the text
I can however, add a `<p>` for the next line before and after the actual info

styles/components/Internship/CommitteeInternshipInfo.module.scss
mattxwang: So you can create a map between committee names and colours; Vincent did this in his PR, here's an example: https://github.com/VeeGee-Creations/website/blob/Reimplement-Navbar/styles/components/Navbar.scss#L130

TL;DR: use `@each` and SASS lists/maps.

components/Internship/InternshipTimeline.js
advaithg: hm i'm actually not sure how to do this
additionally - I think it isn't necessary, since the timeline is purely decorative and not providing any valuable information.
all information in the timeline is also available right underneath it, so making it screen readable would only repeat the information twice (this is also why I decided it was unnecessary on a mobile view)

pages/internship.js
advaithg: would it make sense to redirect to a page saying "coming soon"? I'll ask Ava how she prefers to handle this as well.

advaithg: talked to ava - hiding the button until apps are out!

components/Internship/InternshipTimeline.js
mattxwang: Sure, if that's the case, then add `aria-hidden` or whatever other accessibility annotations are needed.

(the timeline does add information, namely what has "already happened". that being said, that could be inferred by a user, so not too big of a deal)

components/Internship/NextSteps.js
mattxwang: I guess my question is more, can we remove one level of nesting here? There's no styling on `<p>{info}</p>`, so we should be able to make the entire card face (the wrapping `<div>`) a paragraph. There's nothing else in the card face other than the paragraph, right?

advaithg: This would have been applicable a while ago, but in the interests of clarity I've added a heading before the `<p>` with the committee name

data/internship.js
```diff
@@ -0,0 +1,110 @@
...
+            date: '11/06',
+            name: 'Official ACM Onboarding',
+            info: 'Learn what you need to know as an ACM Intern',
+            active: false,
+        },
+    ],
+    testimonials: [
+        {
+            name: 'Maya Raman',
+            quote: <>
```
mattxwang, line 48: All of these should be `<h3>`s in hierarchy. Alternatively, restyle and make the intern name an `<h3>` (semantically makes more sense).

data/internship.js
```diff
@@ -0,0 +1,110 @@
...
+            </>,
+            committee: 'Teach LA',
+            year: '2020-2021',
+        },
+    ],
+    QA: {
+        rows: [
+            {
+                title: 'What is the Internship Program?',
+                content: 'The ACM Internship program is a program that facilitates the transition of UCLA students to ACM officers. Interns learn about the ACM committee they are interning for and ACM in general before becoming an official ACM officer. Although ACM Interns are not paid, they gain valuable experience organizing events or teaching Programming classes or designing event banners and more.',
```
mattxwang, line 81:
```suggestion
                content: 'The ACM Internship program is a program that facilitates the transition of UCLA students to ACM officers. Interns learn about the ACM committee they are interning for and ACM in general before becoming an official ACM officer. Although ACM Interns are not paid, they gain valuable experience organizing events, teaching programming classes, designing event banners, and more.',
```

data/internship.js
```diff
@@ -0,0 +1,110 @@
...
+            },
+            {
+                title: 'What experience do you need to be an intern for ACM?',
+                content: 'None! We are looking for students that are passionate about computer science and are willing to learn. Although a couple intern positions may state a preference for those with experience, no experience is required for any Intern position.',
```
mattxwang, line 85:
```suggestion
                content: 'None! We are looking for students that are passionate about computer science and are willing to learn. Although a couple intern positions may state a preference for those with experience, no experience is required for any intern position.',
```

data/internship.js
```diff
@@ -0,0 +1,110 @@
...
+            },
+            {
+                title: 'How many committees can I apply for?',
+                content: 'You can apply to be an intern for up to 3 committees, but you may only be an intern for one of those committees.',
+            },
+            {
+                title: 'When we apply to be an intern for a committee, do we apply to a specific intern position or just the committee as a whole?',
+                content: 'It depends on the committee. The application for each committee will ask what intern position you are interested in if they want their prospective interns to apply to specific positions. But some committees will allow prospective interns to just apply to the committee as a whole.',
```
mattxwang, line 93:
```suggestion
                content: 'It depends on the committee. The application for each committee will ask what intern position you are interested in if they want their prospective interns to apply to specific positions. Other committees will allow prospective interns to just apply to the committee as a whole.',
```

data/internship.js
```diff
@@ -0,0 +1,110 @@
...
+            },
+            {
+                title: 'How can I get involved in ACM if I am not an intern?',
+                content: 'If you are not an intern, there are plenty of ways to get involved in ACM. You can always attend workshops that different committees are hosting. Subscribe to our Newsletter if you would like weekly updates on when workshops are. Some committees, like Teach LA, allow students to be officers/contributing members without participating in the ACM Internship Program.',
```
mattxwang, line 97:
```suggestion
                content: 'If you are not an intern, there are plenty of ways to get involved in ACM. You can always attend workshops that different committees are hosting. Subscribe to our newsletter if you would like weekly updates on when workshops are. Some committees, like Teach LA, allow students to be officers/contributing members without participating in the ACM Internship Program.',
```

data/internship.js
```diff
@@ -0,0 +1,110 @@
...
+            },
+            {
+                title: 'I have more questions, how can I get them answered?',
+                content: <>If you have any questions about the application process or the Internship program in general email <a href="mailto:ava24@g.ucla.edu">Ava Asmani</a>, the Internship Director. If you have any questions that are specific to committees, there will be an Internship Program ACM Committee Info Session/Q&A sometime in week 3 or 4 of the Fall Quarter.</>,
```
mattxwang, line 101:
```suggestion
                content: <>If you have any questions about the application process or the Internship program in general email <a href="mailto:ava24@g.ucla.edu">Ava Asmani (ava24@g.ucla.edu)</a>, the Internship Director. If you have any questions that are specific to committees, there will be an Internship Program ACM Committee Info Session/Q&A sometime in week 3 or 4 of the Fall Quarter.</>,
```

pages/internship.js
```diff
@@ -0,0 +1,109 @@
...
+import data from '../data';
+
+import styles from '../styles/pages/Internship.module.scss';
+
+const { committees, internship, board } = data;
+
+const internshipInfo = board.concat(committees);
+
+function InternshipPage(){
+    const { items, testimonials, QA, FAQStyles } = internship;
```
mattxwang, line 23: Is there a reason why this is inside the component?

styles/components/Internship/CommitteeInternshipInfo.module.scss
```diff
@@ -0,0 +1,63 @@
...
+  margin: auto;
+  margin-bottom: 50px;
+  padding: 1.125em 1.5em;
+  position: relative;
+  width: 80vw;
+  
+  .internship-info-data {
+    font-size: 0.8em;
+    line-height: 150%;
+    margin: 0 22px 14px;
```
mattxwang, line 18: I personally dislike the three field `margin`, since it's confusing. Can we enumerate all four here?

styles/pages/Internship.module.scss
```diff
@@ -0,0 +1,103 @@
+@import '../variables.scss';
+
+$single-column-breakpoint: 500px;
+
+@mixin shadow-boxes {
+  box-shadow: 0 0.125rem 0.5rem rgba(0, 0, 0, 0.3), 0 0.0625rem 0.125rem rgba(0, 0, 0, 0.2);
+}
```
mattxwang, line 7: Can we put this in globals / variables?

mattxwang (changes requested):
I know we're tight on time, so a lighter review than normal.

First, do the cards work? On Chrome/macOS, I get this:

https://user-images.githubusercontent.com/14893287/134830409-7602f9c2-047b-4438-834e-64ddb85802db.mp4

Not sure what's happening here, but that means that none of the cards work.

Broadly, the FAQ component is also still not tabbable.

Re: content changes, you can commit them as a batch at once.

evanzhong:
https://github.com/uclaacm/website/pull/222/commits/91e04333bd97402c5f854f33f45fcc442267b0b3 along with https://github.com/uclaacm/website/pull/222/commits/67471a133fc576e65b097cb5a8c5fae1462a20d4 is kind of a horrible hack, but they get the job done

pages/internship.js
```diff
@@ -0,0 +1,109 @@
...
+                <div className={'grid-tablet-2'}>
+                    <section>
+                        {items.map((item, i) => (
+                        <div key={i} className={styles['application-item']}>
+                            {item.date} &nbsp;
+                            {item.name}
+                            <p className={styles['application-item-info']}>{item.info}</p>
+                        </div>
+                        ))}
+                        -- more dates will be added throughout the quarter --
```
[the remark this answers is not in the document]
advaithg: I think we can leave this in, especially considering the timeline is blurred - and confirm with ava in the board meeting when finalizing the timeline

data/internship.js
advaithg: im not sure why the name being a h3 makes more sense - since it isn't serving as a header
resetting them to h3 might be a better optio

PR #223, opened 2021-08-01
Renames `.stylelintrc.json` -> to `stylelint.config.js`

### LINKED ISSUE #206: Switch Stylelint config to `.js` file
I think we should switch our config over to a `.js` file, for two key reasons:

1. we can comment
2. we can have lists with a trailing comma dangle, which ensures clean diffs 

This should be as simple as renaming the current config file to `.stylelintrc.js`, and then exporting the top-level object + applying any necessary linting.

mattxwang:
Closes #206. Chose to use `stylelint.config.js` since then it gets hit by ESLint, but fine with using `.stylelintrc.js`, etc. too!


evanzhong (approved):
lgtm,

Really appreciate this change, as mentioned in #206 this allows for cleaner diffs

PR #224, opened 2021-08-01
disables `@next/no-img-element` in relevant areas, and adds `TODO` comments where needed

mattxwang:
Resolving our issues with `next/image` is already on the roadmap with #188. In the meantime, as pointed out in #216, we have some noisy errors that don't particularly help us. This PR disables rules *just for those lines*!


evanzhong (approved):
lgtm

PR #229, opened 2021-08-04
Fixed Page nav-bar location hash UI bug #228

### LINKED ISSUE #228: Committee Page nav-bar location hash UI bug
When navigating around the committees page we use location hashes. For example navigating to `/committees#studio` shows the following, with the committee graphic slightly occluded by the nav bar
image

This issue likely stems from the the way we style our nav-bar. Because we used fixed positioning (`position: fixed;`) the location hash based autoscroll clips part of the heading image.
image

krishrahul98:
Closes #228.

### DIFF
--- a/styles/pages/committees.scss
+++ b/styles/pages/committees.scss
@@ -76,3 +76,10 @@
     }
   }
 }
+
+// this is for fixing location hashes scroll bug
+
+html {
+  scroll-padding-top: 75px;
+}
+


mattxwang (approved):
Did some local testing, and works for me! Thanks for the contribution @krishrahul98.

(if/when we split up the committees page to use CSS modules, we'll probably want to move this to the global style)

krishrahul98:
@mattxwang yes sure

PR #236, opened 2021-08-11
rework footer: overhaul CSS, link to all pages on site

mattxwang:
This PR does two main things:

1. swaps out the footer CSS to be module'd, use grid in favor of fixed-width spacing, and uses classes instead of nested specificity
    * this comes with the added benefit of increasing accessibility, since we're relying on less hacky positioning!
2. adds links to the footer - every single page under the "ACM" category, and a link to each committee
    * under evan's suggestion, I've converted each committee link to a dark version of their logo + wordmark, to add some flair!

Motivating goals for adding the links to the footer:

1. as we add new pages (Impact, JEDI, etc.) we need to put them somewhere - and the navbar is quite a bit of work!
2. for the Tech Gala page, there's no way to get to it other than the homepage alert banner
3. the current flow to get to a committee's website is always 2+ clicks from the homepage, which seems bad 
4. this is typical practice (I personally always scroll down to footers to find relevant links)

Some comments I'll make:

* ~~I kinda dislike how the links align, but a simple left/right text align looks even worse~~
* ~~how often do the footer links change? we may want to plop this into a data file~~
* when we add JEDI/Impact/w.e, where should they go? I'm thinking a new section under ACM called "Events and Initiatives", and we can bundle tech gala in there as well
* ~~are there any spacing issues and/or mobile view things I should look into fixing?~~
* added an extra util

Inadvertently part of #191.


components/Footer.js
```diff
@@ -1,40 +1,86 @@
...
+const footerACMLinks = [
+	{ title: 'About', path: '/about' },
+	{ title: 'Committees', path: '/committees' },
+	{ title: 'Events', path: '/events' },
+	{ title: 'Sponsors', path: '/sponsors' },
+	{ title: 'Tech Gala', path: '/techgala' },
+	{ title: 'Membership Portal', path: 'https://members.uclaacm.com', ext: true },
+];
+
+// Do we want to prepend ACM here?
```
evanzhong, line 16: My preference would be to add the ACM part here

components/Footer.js
```diff
@@ -1,40 +1,86 @@
...
+const footerCommitteeLinks = [
+	{ title: 'AI', path: 'https://uclaacmai.github.io/', ext: true },
+	{ title: 'Cyber', path: 'https://acmcyber.com', ext: true },
+	{ title: 'Design', path: '/committees#design' },
+	{ title: 'Hack', path: 'https://hack.uclaacm.com', ext: true },
+	{ title: 'ICPC', path: 'https://icpc.uclaacm.com', ext: true },
+	{ title: 'Studio', path: 'http://localhost:3000/committees#studio' },
+	{ title: 'Teach LA', path: 'https://teachla.uclaacm.com', ext: true },
+	{ title: 'W', path: '/committees#w' },
```
evanzhong, line 25: How did you settle on these links? If I recall correctly, we have some committee data stored in objects in `/data/committees.js`. 

Do we want to standardize the links we pull somehow. The nav gets a little tricky with the footer committee links since many of lead to external sites while the links in our `CommitteeSpread` component on the landing page all route to the `/committee` page.

components/Footer.js
```diff
@@ -1,40 +1,86 @@
...
+];
+
+// Do we want to prepend ACM here?
+const footerCommitteeLinks = [
+	{ title: 'AI', path: 'https://uclaacmai.github.io/', ext: true },
+	{ title: 'Cyber', path: 'https://acmcyber.com', ext: true },
+	{ title: 'Design', path: '/committees#design' },
+	{ title: 'Hack', path: 'https://hack.uclaacm.com', ext: true },
+	{ title: 'ICPC', path: 'https://icpc.uclaacm.com', ext: true },
+	{ title: 'Studio', path: 'http://localhost:3000/committees#studio' },
```
evanzhong, line 23:
```suggestion
	{ title: 'Studio', path: '/committees#studio' },
```

components/Footer.js
```diff
@@ -1,40 +1,86 @@
 import Link from 'next/link';
 import React from 'react';
+import styles from '../styles/components/Footer.module.scss';
 import SocialMedia from './SocialMedia';
 
+// TODO: add GM?
```
evanzhong, line 6: works for me!

components/Footer.js
```diff
@@ -1,40 +1,86 @@
...
+	{ title: 'Studio', path: 'http://localhost:3000/committees#studio' },
+	{ title: 'Teach LA', path: 'https://teachla.uclaacm.com', ext: true },
+	{ title: 'W', path: '/committees#w' },
+];
+
+function FooterLinkElement({ title, path, ext }){
+	return (
+		<li>
+			<Link href={path}>
+				<a className={styles['link-footer']} target={ext ? 'target': ''} rel={ext ? 'noopener noreferrer' : ''}>{title}</a>
```
evanzhong, line 32: I've only seen `target="_blank"` before, but verified that this works locally so lgtm

components/Footer.js
```diff
@@ -1,40 +1,86 @@
...
-						<h3>Reach us at</h3>
-						<a href="mailto: acm@ucla.edu" className="email"><span className="footer-text">acm@ucla.edu</span></a>
+					<div>
+						<h3 className={styles['footer-header']}>About ACM at UCLA</h3>
+						<ul className='list-unstyled'>
+							{
+								footerACMLinks.map((link) => <FooterLinkElement key={link.path} {...link} />)
+							}
+						</ul>
+						{/* TODO: consider where to put impact/jedi! events & initiatives? */}
```
evanzhong, line 64: Could just add another `<h3>` "Initiatives" (?) and render them here in the same block

evanzhong (changes requested):
Love the use of css-grid! Most of my review is left in comments.

Looking good! A couples things of note:
- Link routing consistency. When we say a link routes to `ACM AI` what do we mean? Does it link to the AI section under our `/committees` page or AI's own external site? Potential solution could be leaning on our `/data/committees.js` file
- Headers on mobile (`About ACM at UCLA`, `Committees`) might need some adjusting since they're just hovering the middle right now. Might be good to boost the font size too so the hierarchy is more apparent image
- The content alignment doesn't really bother me for the `About ACM at UCLA` section but I do agree the `Committees` section looks a bit odd. A potential solution would be to use a row of committee diamonds (or even a column?) since those are uniform size.
  - There's enough space in the footer to accommodate a row of diamonds, as a row of logos is what we do for the social media stuff image
  - This may even help the whole footer look more balanced?
  - Downside is this is a textless solution so we'd be relying on the diamonds to convey full meaning of committees. (You _could_ have each committee's name under the diamond though?)

components/Footer.js
mattxwang: Ooh great catch, it should be `target="_blank"` - my bad!

components/Footer.js
mattxwang: Yup, I'm down for that too!

components/Footer.js
mattxwang: Cool, we can do it once GM is reenabled.

mattxwang:
> Link routing consistency. When we say a link routes to ACM AI what do we mean? Does it link to the AI section under our `/committees` page or AI's own external site? Potential solution could be leaning on our `/data/committees.js` file

> How did you settle on these links? If I recall correctly, we have some committee data stored in objects in `/data/committees.js`. 

> Do we want to standardize the links we pull somehow. The nav gets a little tricky with the footer committee links since many of lead to external sites while the links in our `CommitteeSpread` component on the landing page all route to the `/committee` page.

Yeah, these are all great questions. I do think I'd like to link to their external sites, if possible - I imagine they'll be more frequently updated than the committees page itself, and might be more useful for a user hunting for "ACM AI" (opposed to just learning more about committees). That being said, if there's a strong reason to not do so, more than happy to entertain it.

In terms of technical implementation, one way we can try to use `/data/committees.js` is to add two direct fields to it:

* an optional `external_site` (or similarly-named) variable; the advantage of this over the current approach is we can directly access it without looping through the `links` array
* `route` (or `shortname`, `slug`, etc.) - this tells us the local `/` path. should likely replace the `class` variable there now

Then, we first check if `external_site` exists; if it does, let's use it, and if it doesn't, use `/route` instead. Thoughts?

mattxwang:
> Headers on mobile (`About ACM at UCLA`, `Committees`) might need some adjusting since they're just hovering the middle right now. Might be good to boost the font size too so the hierarchy is more apparent

Gave it a shot - let me know! I didn't boost too much of the font size to keep the original vibe from live, but more than happy to change that since I think visibility/hierarchy is the most important!

mattxwang:
> The content alignment doesn't really bother me for the `About ACM at UCLA` section but I do agree the `Committees` section looks a bit odd. A potential solution would be to use a row of committee diamonds (or even a column?) since those are uniform size.

Hm, that's an interesting idea - I'm a bit wary on fully relying on images (as you've said) since I'm not sure if the diamonds enough are explanatory enough of what the committee is, i.e. the user can't just scan and see what it is. What are your thoughts on doing something like the committees page sidenav? A bit frustrating in its own right since we'd need to load a different set of wordmarks for a dark background, but willing to give it a shot!

evanzhong:
> > Headers on mobile (`About ACM at UCLA`, `Committees`) might need some adjusting since they're just hovering the middle right now. Might be good to boost the font size too so the hierarchy is more apparent
> 
> Gave it a shot - let me know! I didn't boost too much of the font size to keep the original vibe from live, but more than happy to change that since I think visibility/hierarchy is the most important!

lgtm, it's less ambiguous now
image

evanzhong:
> In terms of technical implementation, one way we can try to use `/data/committees.js` is to add two direct fields to it:
> 
> * an optional `external_site` (or similarly-named) variable; the advantage of this over the current approach is we can directly access it without looping through the `links` array
> * `route` (or `shortname`, `slug`, etc.) - this tells us the local `/` path. should likely replace the `class` variable there now
> 
> Then, we first check if `external_site` exists; if it does, let's use it, and if it doesn't, use `/route` instead. Thoughts?

I like the idea of an `external_site` as well as a `route` variable. 

Though this does bring up the structure of the `links` array. Can we const-ify the platforms? Is there something more specific about the naming of the variable? Might warrant a separate issue out of the scope of this PR

evanzhong:
> 
> Hm, that's an interesting idea - I'm a bit wary on fully relying on images (as you've said) since I'm not sure if the diamonds enough are explanatory enough of what the committee is, i.e. the user can't just scan and see what it is. Don't think labelling each diamond works properly ("Teach LA" and "W" are so different in length). What are your thoughts on doing something like the committees page sidenav? A bit frustrating in its own right since we'd need to load a different set of wordmarks for a dark background, but willing to give it a shot!

Something like the committee nav would be nice, and I think would definitely make the footer look more balanced. Is there no way you can repurpose the `/Committees/Sidebar` component that already exists? Maybe add a boolean prop `darkMode` defaulting to false?

mattxwang:
> > In terms of technical implementation, one way we can try to use `/data/committees.js` is to add two direct fields to it:
> > 
> > * an optional `external_site` (or similarly-named) variable; the advantage of this over the current approach is we can directly access it without looping through the `links` array
> > * `route` (or `shortname`, `slug`, etc.) - this tells us the local `/` path. should likely replace the `class` variable there now
> > 
> > Then, we first check if `external_site` exists; if it does, let's use it, and if it doesn't, use `/route` instead. Thoughts?
> 
> I like the idea of an `external_site` as well as a `route` variable.
> 
> Though this does bring up the structure of the `links` array. Can we const-ify the platforms? Is there something more specific about the naming of the variable? Might warrant a separate issue out of the scope of this PR

Just wrote a short fix for this, introducing a new variable called `slug` (similar to `route`). If we proceed w/ this approach, I'll later write another PR that swaps out all uses of `class` as variable to `slug`.

mattxwang:
> Something like the committee nav would be nice, and I think would definitely make the footer look more balanced. Is there no way you can repurpose the `/Committees/Sidebar` component that already exists? Maybe add a boolean prop `darkMode` defaulting to false?

Hm, repurposing the existing component would be a bit tricky since the link that we pick would have to change depending on the committee (for the footer, but not for the committees page). I can def work on that and/or just writing a new component - would you prefer for me to do it in this PR, or follow this up with another one later?

data/committees.js
```diff
@@ -4,6 +4,7 @@ export default [
 	{
 		name: 'Studio',
 		class: 'studio',
+		slug: 'studio',
 		tagline: 'Creative Expression through Games and VR',
```
evanzhong, line 8: For the `external_link` variable, thoughts on explicitly writing: `external_link: null`, so the fields are consistent for all committees?

I guess this really depends on if every committee wants an external site / will eventually create one

evanzhong (approved):
Nice work on the `slug` field. Left one minor comment

Couple of things
- I agree data sanitization for the `committees.js` data file can be done in a separate PR
- I'd prefer changes to the committee links in the footer be done in this PR if possible. 
  - I think it would be good to avoid commit pollution if we're just co-opting the existing component or making a very similar one.
  - Additionally I think it would be good for the creation (or refactor) be tied to the forcing function in our commit history, thus would make sense for it to ship in this PR.
  - However: If this feature is a super high priority item I am okay with shipping it as is, thus the approving review.

mattxwang:
> I'd prefer changes to the committee links in the footer be done in this PR if possible.

Sounds good to me, will work on it tomorrow/day after!

data/committees.js
mattxwang: I don't mind, especially if it makes things clearer! Will address tomorrow!

mattxwang:
@evanzhong just implemented the footer w/ dark wordmarks, let me know what you think.

Also wrapped up some accessibility stuff, that long with Vincent's changes, addresses #247 / the long-standing accessibility issues. I took a different approach (i.e. by adding a screenreader-only header that outlines the footer).

evanzhong (approved):
lgtm!

Couple minor things: 
- Have we gotten a QA from design for the dark thumbnails? image
  - Would this look cleaner if we only used the workmarks?
- Could you make that change to make `external_link` to explicitly `null` for committees that don't have an external site?

PR #237, opened 2021-08-11
updates next to `v11.1.0`, removes postcss as a dev dependency

### LINKED ISSUE #186: remove postcss@8 as a dev dependency
RIght now, this dependency exists as a stopgap to fix [this](https://github.com/vercel/next.js/issues/24882) error in next.js

To fix this, you'll probably have to resolve this error and then remove the dependency.

mattxwang:
This PR manually updates next to `v11.1.0`; there are two good reasons to do this:

1. [it resolves a security vulnerability](https://github.com/vercel/next.js/releases/tag/v11.1.0) (though we don't have a custom error page)
2. it lets us close #186 as it merges the PR that resolves the issue upstream

(note: I totally messed up from the CLI and accidentally pushed to `main`; even without `--force`, it succeeded. seems a bit sus; sorry for the commit pollution!)

Closes #186.


advaithg (approved):
Looks good! Doesn't break anything and deals with one of our post next transition issues!

Issue #242, opened 2021-08-17
Double corner borders on techgala page


advaithg:
<img width="822" alt="screenshot of project 'Am I Busy?' from the techgala page" src="https://user-images.githubusercontent.com/67539191/129655609-9d779459-bffb-471c-a358-18419d259e24.png">

On some of the projects on the tech gala page, a double border is visible in the corners - this occurs on all the project images, but it's only visible on the ones where the background is light enough to see it.

We've run into this issue before with the committees page, so I'll probably deal with this myself! It's to do with the CSS on the divs surrounding `next/image`, so it needs an override to fix.

PR #243, opened 2021-08-17
Fixes techgala page image border issues

advaithg:
Resolves issue #242 

Removes the borders around the project images entirely using the same overrides used in the committees page


evanzhong (approved):
Ah we meet again at last, `!important`!

While I'm not a fan of using `!important` to override CSS rules, this is just personal preference and there's nothing blocking this from shipping. I generally try to avoid using it as it disrupts the cascading nature of CSS but sometimes it's the easiest way to achieve some layouts, as might be the case right here.

PR #238, opened 2021-08-12
Impact Page

mattxwang:
[very draft]

https://deploy-preview-238--jovial-pasteur-581b4a.netlify.app/impact

PR #235, opened 2021-08-11
JEDI Page

mattxwang:
This PR implements three pages from [Helia's wireframe](https://docs.google.com/presentation/d/1dEaIW6uVE0LcGbXHKtroNc5um3K8sHQkp6WxQGpZWj0/edit#slide=id.gd7e5a0907d_0_18):

* [JEDI home page](https://deploy-preview-235--jovial-pasteur-581b4a.netlify.app/jedi)
* [meet the JEDIs](https://deploy-preview-235--jovial-pasteur-581b4a.netlify.app/jedi/meet)
* [allyship spaces](https://deploy-preview-235--jovial-pasteur-581b4a.netlify.app/jedi/allyship)

Outside of directly implementing the pages, we:

* add an "initiatives" section to the footer
* imports more font weights for `Open Sans`, includes semi-bold, bold, and bold italic

Technical to-do tracker:

* general
  - [x] SEO
  - [x] mobile responsiveness
  - [x] screenreader audit
  - [x] implement nav?
  - [x] add to footer
  - [x] `<a>` and button CSS
  - [ ] resolving #188 
  - [ ] homepage above-the-fold image
  - [x] adding more color/holograph effect
  - [x] implement contact feature
* allyship spaces page
  - [x] move data to its own file
  - [x] move allyship card to its own component
  - [ ] scope module css to just the allyship card
  - [x] decide on impl. method for links
  - [x] ~~icons for link buttons!~~ going to punt this as out of scope unless Design requires it for QA
  - [x] ~~programmatic dates~~ going to punt this as out of scope 
  - [x] ~~location links?~~ going to punt this as out of scope 
* meet the jedis page
  - [x] move data to its own file
  - [x] move jedi card to its own component
  - [ ] scope module css to just the jedi card
  - [x] proper styling of pronouns
  - [x] font size check
  - [ ] stretch goal: center orphaned grid elements
  - [ ] stretch goal: change gutter spacing to be fixed-gutter, variable image

PR #244, opened 2021-08-18
per-page SEO (meta tags, opengraph) via `next-seo`

### LINKED ISSUE #116: Populate page metadata, social media metadata

### LINKED ISSUE #194: Link embeds are flaky
Just did some testing with our new Next site, and noticed that link embeds are still a flaky. Some notes on the homepage (likely the most important page):

* on Discord, there's no embed at all
* on facebook messenger, the image doesn't work as intended
* works on Instagram DM, though the image is the first news item
* works on FB post, though the image is the first news item

This should be part of a larger issue surrounding one of the post-action items of #178, which is properly filling out SEO - i.e. #116.

mattxwang:
This PR implements per-page SEO tags via [`next-seo`](https://github.com/garmeeh/next-seo). Under-the-hood, the package handles basic meta tags, and OpenGraph URLs; according to the documentation, Twitter now also parses OG data, and thus doesn't generate twitter-specific metadata.

I then manually write and apply copy for each page: the home page, about, committees, events, sponsors, and tech gala. We'll have to backport this to other new pages, like the ones introduced in #238 and #235. I chose not to add any default templating or abstract this to a component, since I think the copy differs enough where boilerplate doesn't help us too much.

For reviewers - it would be great if you could test this on a variety of social media platforms (Facebook, Twitter, Discord, Instagram. etc.). I've done Discord testing in the #test3 moderator channel of the server, as well as DMing myself on FB messenger.

Closes #116, (hopefully) closes #194.

### DIFF
--- a/pages/about.js
+++ b/pages/about.js
@@ -1,3 +1,4 @@
+import { NextSeo } from 'next-seo';
 import Image from 'next/image';
 import React from 'react';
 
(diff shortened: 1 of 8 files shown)


mattxwang:
Oh also, this is probably easier to review w/ whitespace changes hidden.

pages/about.js
```diff
@@ -17,75 +18,88 @@ function About() {
...
-				</div>
-				<div className={styles['text-section']}>
-					<h1>What are ACM committees?</h1>
-					<p>
-						ACM is comprised of eight committees — each serving a different topic and mission.
-						We strive to cover a plethora of interests and encourage members to explore new topics, too!
-					</p>
+			<NextSeo
+				title="About | ACM at UCLA"
+				description="As a student chapter of the international Association for Computing Machinery,</a> ACM at UCLA is the largest Computer Science student organization at UCLA and in Southern California. We welcome students of all backgrounds and skill levels to join our community and share our love for technology!"
```
evanzhong, line 23: errant closing </a> tag?

evanzhong (approved):
Really great stuff. No huge blockers from merging, I think there's a small typo to fix but that's about it.

pages/about.js
mattxwang: Haha, yes - great catch!

Issue #218, opened 2021-07-26
Long-term: CI accessibility audits


mattxwang:
(matt will flesh this out later!)

mattxwang:
Also, just wanted to shout out Tim Gu's detailed issue in the hack repo:  https://github.com/uclaacm/hack.uclaacm.com/issues/229

mattxwang:
aXe might be a good tool; look at:

* https://medium.com/induction/access-your-accessibility-installing-and-running-axe-with-nextjs-f06e23378df2
* https://web.dev/accessibility-auditing-react/

mattxwang:
This should be easier now that #245 is merged!

mattxwang:
Something that Arjun sent my way: https://userway.org/

mattxwang:
https://kittygiraudel.com/2020/12/01/a11y-advent-calendar/

mattxwang:
https://github.com/nickcolley/jest-axe

PR #245, opened 2021-08-18
starts automated accessibility steps: loads `@axe-core/react` on `next dev`, enables `eslint-plugin-jsx-a11y` on `strict`

mattxwang:
This PR does a couple of things:

1. installs and enables `eslint-plugin-jsx-a11y`, and turns it on with `strict` mode
    * this triggers on the navbar improperly using the `<label>` element, which is a valid concern. reworking it involves changing the DOM structure of the navbar, which in turn breaks the CSS - I'd like to table fixing this until we rework the navbar
    * this also triggers on next's `<Link><a></a></Link>` pattern, which is realistically a false-positive. Unfortunately, there's no great solution to this, and as such I've temporarily disabled the rule for `<Link>`; see the comments in the eslint config for more info
2. installs and enables `@axe-core/react`, which logs accessibility issues to the console in dev
    * there are **a ton** of issues to resolve. so, I'll scope that out for another PR instead. 

Something that's still left to be desired is to get aXe's level of accessibility specificity, but getting it as a CI action (like `eslint-plugin-jsx-a11y`) - will figure that out soon!

Part of #218.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,6 +8,7 @@ module.exports = {
     'eslint:recommended',
     'plugin:react/recommended',
     'plugin:@next/next/recommended',
+    'plugin:jsx-a11y/strict',
   ],
   parserOptions: {
     ecmaFeatures: {
@@ -16,7 +17,11 @@ module.exports = {
     sourceType: 'module',
   },
   ignorePatterns: ['/build/*'],
-  plugins: ['react', 'import'],
+  plugins: [
+    'react',
+    'import',
+    'jsx-a11y',
+  ],
   settings: {
     react: {
       version: 'detect',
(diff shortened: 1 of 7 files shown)


evanzhong (approved):
lgtm.

Small thing:
- >this triggers on the navbar improperly using the <label> element, which is a valid concern. reworking it involves changing the DOM structure of the navbar, which in turn breaks the CSS
  - I think tabling this rather large change is the right call here. If you haven't already, could you create a new issue outlining a navbar refactor and reference this PR.

mattxwang:
Huh, having some problems w/ the merge - let me roll this back.

mattxwang:
> I think tabling this rather large change is the right call here. If you haven't already, could you create a new issue outlining a navbar refactor and reference this PR.

Yup, will do!

PR #249, opened 2021-08-20
resolves axe errors in #247

VeeGee-Creations:
### Resolve Secondary Font Color Contrast

variables.scss:
* font-secondary-color: #777 fails contrast on most text sizes with a contrast of 4.47:1 on white background
* darkened font-secondary-color: #595959 passes all text sizes with a contrast of 7:1 on white background

### Resolve `<ul>` Nesting

Navbar.js:
* wrapped .committee-mobile-nav in `<li>`
* gave `<li>` id committee-mobile-nav-item

Navbar.scss:
* targeted #committee-movile-nav-item to fit height to content

### Resolve Landmark Error

Layout.js:
* added 'main' role to .main-container

Navbar.js:
* added 'banner' role to #navbar
* added 'navigation role to #desktop-nav and #hamburger-menu

Issue #247, opened 2021-08-20
Resolve outstanding `axe` issues!


mattxwang:
After #245, we've now implemented `axe` into our development workflow, and it logs sets of errors to the console. We have an *insane* amount of errors across our pages. Slowly but surely, we should knock them out.

This is a great learning experience re: web accessibility! Might warrant a good first issue, or a walkthrough with the dev team.

mattxwang:
I believe we've covered the initial issues we found from axe. We'll continually have to monitor this (I'd like to CI it, if possible?) as we add more pages and make style changes to the site.

PR #250, opened 2021-08-21
resolves more axe errors

mattxwang:
Now that #249, there are only a couple more axe errors on our radar to resolve. This PR takes the low-hanging fruit;

* we had used `#sign-up-section` in multiple different components that appeared on the same page. to resolve this in a low-impact manner, we shift the styling to use classes instead; we also remove an unstyled `#sign-up-left`
* the events page did not have a `<h1>` tag, so change the main header to use it!

There's a few remaining general issues with accessibility, namely the inaccessibility of our navigation - which is partially flagged by axe (the footer incorrectly uses headers). I'll resolve this in #236!

related to (but does not close): #247.

### DIFF
--- a/components/TechGala/ViewBanner.js
+++ b/components/TechGala/ViewBanner.js
@@ -4,7 +4,7 @@ import React from 'react';
 function TGBanner() {
     return (
         <div className="full-width blue">
-            <div id="sign-up-section" className="banner-section">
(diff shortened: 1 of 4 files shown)


advaithg (approved):
Looks good! Great progress on dealing with accessibility issues!

Issue #225, opened 2021-08-02
Clean up accessibility tree


mattxwang:
Like many of my ideas for this website, this one comes from poking around Hack's (see uclaacm/hack.uclaacm.com#229) - our accessibility tree is rather disgusting.

<img width="562" alt="Screen Shot 2021-08-02 at 10 06 47 AM" src="https://user-images.githubusercontent.com/14893287/127898460-c6a04029-51d1-4009-95a7-5693bf5b85cb.png">

In particular - the nav is poorly marked, many alt tags are not helpful, and the page is lacking a parseable structure.

Debating if this is a walkthrough I'll do with the dev team!

mattxwang:
Some helpful resources: 

https://developer.chrome.com/docs/devtools/accessibility/reference/

https://developers.google.com/web/fundamentals/accessibility/semantics-builtin/the-accessibility-tree

mattxwang:
This should be easier now that #245 is merged!

PR #251, opened 2021-08-21
Axe Errors Continued

VeeGee-Creations:
### Resolved Header Hierarchy
events.js:
* changed h2 to h1 to resolve issue of no h1

Footer.js:
* changed both h3 to h2 to resolve new error of headers skipping levels

Footer.scss:
* pointed h3 styling at the new h2

After searching every page of the site, this should be the last of the current Axe errors.

PR #252, opened 2021-08-21
Accessibility Navbar as per #225

VeeGee-Creations:
Navbar.js:
* Logo link to home now ARIA labeled
* ul elements given presentation role for better screen reading
* hamburger now tab indexed for screen readers
* hamburger given role of button with expanded attribute changed by current state of checkbox

Navbar.scss:
* changed ul.nav-items li from opacity to display
  * no transitions were in use
  * display: none / display: block hides menu from screen reader when menu collapsed

### DIFF
--- a/components/Footer.js
+++ b/components/Footer.js
@@ -8,11 +8,11 @@ function Footer(){
 			<div id="footer-inner">
 				<div id="footer-items">
 					<div className="footer-item">
-						<h3>Find us on social media</h3>
+						<h2>Find us on social media</h2>
 						<SocialMedia type="light"/>
 					</div>
 					<div className="footer-item">
-						<h3>Reach us at</h3>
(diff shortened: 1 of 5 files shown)


mattxwang (changes requested):
Hey Vincent,

Thanks again for tackling on another issue! I can give a more in-depth code review in a bit, but just two general questions:

1. How do the changes between this PR differ from #251 and #253? I see a bit of overlapping code; are you looking to merge them in a particular order?
2. Do you mind trying to silo changes into each PR? For example, this one also edits the Footer - it'll be easy for us to individually provide feedback, merge, and resolve merge conflicts if the PRs are relatively different. For this one, for example, the Footer changes clash with #236 (which is not your fault at all, as it's not immediately obvious!)

PR #272, opened 2021-09-18
new events page

### LINKED ISSUE #115: Long-term: Reworked Events Page

mattxwang:
This PR tackles the MVP for #115. It implements a first-pass of the Events page that Angela designed (shoutout Angela!), by using [Jason Quense's React Calendar](https://github.com/jquense/react-big-calendar) with the [moment localizer](https://momentjs.com/). To modify the styling to match Angela's design, we override (with `!important`, unfortunately) the existing class structure.

In addition to dropping the calendar in, I've also built an aside/side selector that displays more information about the selected event (when going from month/week/day view, but not agenda - this is a shortcoming). 

Both the calendar and the aside view are powered by a new `data/event.js`. The data structure of an individual event is highlighted in #281. While this data could be implemented by hand, that's a pain; I've used `csv-parser` to implement two scripts, `reg` and `seg`, to convert marketing spreadsheet info for recurring and single-time events respectively.

On the bright side, this is a very decoupled design. In particular, if we do end up building an API for our events, we could easily hook this in to the calendar with minimal changes required!

## Some notes and things that should be wrapped up later

* on the calendar itself
    * some of the other views are what I would call "low in design quality", since there is no corresponding mockup. the agenda view is particularly bad.
    * similarly, mobile view is subpar
    * there are contrast issues with the white text on the gradient
    * in general, I don't think this calendar is very accessible, though I haven't had much time to test.
* on the event structure
    * currently there is no particular support for virtual/in-person events, or highlighting things like zoom links, feedback forms, or resources. we didn't have this in the design, but might be good to do!
* on the script
    * no support for image parsing or alt text retrieval; this is rather complicated given our technical constraints (see #281)
    * link support is hard-coded; there is no support for Zoom, IG, Itch, Discord, Google Forms, GitHub, etc.
    * no support for all-day events
    * the generated description does not apply link effects or any markdown-esque styling
    * currently, there is no mechanism to merge the output of the script with the existing input; this is going to be a headache down the line
    * id `offset` is hard-coded
    * little field validation is done
    * the recurring week range requires a slightly different column format than what we have now - `2021-10-04 - 2021-12-01` instead of relying on the quarter system
    * lots of things are hard-coded, so we'll need to keep an eye out for minor human errors

## Usage and Testing

To test: https://deploy-preview-272--jovial-pasteur-581b4a.netlify.app/events

To run the relevant scripts:

```bash
# convert a single event spreadsheet to output.json
$ npm run seg -- input.csv
...
# convert a recurring event spreadsheet to output.json
$ npm run reg -- input.csv
...
```

You'll then need to take the output of `output.json` and put it into `data/events.js` - there's no great way to circumvent this without doing some metaprogramming :) 

**Note:** if you're pulling the CSV from the marketing spreadsheet, **please remove the first row**; the second row is where the headers start!

**Note:** the recurring week range requires a slightly different column format than what we have now - `2021-10-04 - 2021-12-01` instead of relying on the quarter system.

Closes #115 and #281.


mattxwang:
Not sure why the Netlify deploy preview isn't working, going to troubleshoot that later.

mattxwang:
Tagging in @evanzhong; I think other than the event dates, we're in a relatively good spot.

PR #307, opened 2021-10-05
Updates Internship Page

### LINKED ISSUE #293: internship page - make FAQ tabbable
The FAQ element at the end of the internship page is currently not tabbable, which is a big no in terms of accessibility.

To fix this, you'd probably have to
- rip out the component currently used
- either write your own FAQ component, or use an accessible accordion like [this one](https://www.npmjs.com/package/react-accessible-accordion).

advaithg:
Resolves #293 and updates the internship timeline


mattxwang (changes requested):
Generally LGTM.

~~However, you're on the wrong version of node, which is why the `package-lock` has such a large diff. Can you stay on Node 14 for now?~~

~~(in addition, I would expect to see some removals as we're removing a component - are you updating the `package-lock` when you remove the component?)~~

Can you investigate why we're generating a `v2` of `package-lock.json` then?

advaithg:
> However, you're on the wrong version of node, which is why the `package-lock` has such a large diff. Can you stay on Node 14 for now?

`node --version` shows 14.16.1, meaning I'm still on the same version

> (in addition, I would expect to see some removals as we're removing a component - are you updating the `package-lock` when you remove the component?)

`cat package-lock.json | grep "faq"` shows no results, meaning the component got removed. The diff also has 14 deletions.

mattxwang:
As an update, you'll have to de-dupe your data changes from #308.

Issue #351, opened 2021-11-16
Add "initiatives" section to mobile navbar


mattxwang:
This is what the current mobile nav looks like:

![Screen Shot 2021-11-15 at 9 30 35 PM](https://user-images.githubusercontent.com/14893287/141926587-3125d4c3-ac19-4f8a-83a0-babd0a45e592.png)

I'd like to add links to the [impact](https://acm.cs.ucla.edu/impact) and [JEDI](https://acm.cs.ucla.edu/jedi) pages, under a header called "initiatives". Styling-wise, it should operate similar to the committees page, with the caveat that we want to highlight the current page if it's active, not the current committee based off of the hash.

I would start by looking at [`components/Navbar.js`](https://github.com/uclaacm/website/blob/main/components/Navbar.js)!

mattxwang:
@reginawang3495

reginawang3495:
> with the caveat that we want to highlight the current page if it's active, not the current committee based off of the hash.

I don’t understand

mattxwang:
> I don’t understand

If we're on the "JEDI" page, we should highlight the link to the JEDI page; same for Impact.

Marob-nod:
Hi ! Can I work on it ?

Marob-nod:
Hi @reginawang3495 , I didnt know and I just did it. It is working ! :) 
Tell me if you  want me to pull a request.
![Capture d’écran du 2021-11-16 14-48-28](https://user-images.githubusercontent.com/79056820/141997212-2e167425-41d8-4ab9-a255-c4f2260d3b60.png)

reginawang3495:
Sure, if you've already made it then that's fine, send over the pull request :)

reginawang3495:
@Marob-nod any updates? :)

PR #353, opened 2021-11-17
Add navbar mobile initiatives section 

Marob-nod:
In the new "initiatives" section, the links to JEDI and impact pages are working. When it is active, the link is highlighted with the color blue (amc cobalt). 
By default, there is no Link on the button "Initiatives".

PS: Tell me if you want more/other styles

It was a pleasure to work for this great project ! I hope there will be more easy issues to fix ! :)

Capture d’écran du 2021-11-16 14-48-28


mattxwang (changes requested):
Hi @Marob-nod,

Thanks for your contribution! Two quick asks:

1. It looks like you've added a trailing space, and thus your change doesn't pass lint/doesn't build. Could you resolve that? This is autofixable, so `npm run lint-fix` should do the trick.
2. I've noticed you've added `node` as a dependency to the project; why is that? We aren't directly consuming `node`.

Marob-nod:
Hey @mattxwang , I am taking care of it.

Marob-nod:
I ran lint-fix successfully, removed node from the dependencies and finally.
Have a good day :)

components/Navbar.js
```diff
@@ -45,6 +45,13 @@ export default function Navbar() {
                   <li><ActiveLink activeClassName="active" href="/committees#hack" passHref={true}><button type="button" onClick={menuActivate}>Hack</button></ActiveLink></li>
                 </ul>
               </li>
+              <li><button>initiatives</button></li>
```
mattxwang, line 48: Since this isn't clickable, can we use a different element instead of `<button>`? This will trip up a screenreader/makes our nav not as accessible.

mattxwang (changes requested):
Thanks for the quick turnaround! Have one last comment, and then I think this is good to merge!

components/Navbar.js
Marob-nod: No problem. Now, there is just the <li>. Do you think replacing by another element is accurate ?

Btw, "npm run lint-fix" shows several errors but not relating to my work i think.

mattxwang: > No problem. Now, there is just the li. Do you think replacing it by another element is accurate ?

Perfect, seems good to me!

> Btw, "npm run lint-fix" shows several errors but not relating to my work I think.

Interesting, when I run `npm run lint` off of `main`, I have no problems. This also passes CI, so I wouldn't overstress!

mattxwang (approved):
Overall, great work @Marob-nod. This looks good to merge. Thank you for your contribution!

mattxwang:
Closes #351.

Issue #339, opened 2021-11-07
Timezone-safe "countdown"-style element


mattxwang:
As I write more pages that involve showing content for a specific event, I find myself wanting to be able to show content _after_ a specific time has passed (think a countdown of sorts). I think this is an opportunity for us to write a very helpful wrapper component! I'm thinking of a component that:

1. Has a mandatory prop that represents time somehow (UNIX timestamp? JS timestamp? `moment`?)
2. Has a `placeholder` or `fallback` component prop that is rendered if the time hasn't happened yet
3. Renders its children if the time has happened yet

The advantage here is that we don't have to rewrite time-checking logic (which is notoriously frustrating), and instead provide our developers with a high-level API to toggle content.

I see this being useful for:

* town hall (the driving motivator for writing this ticket)
* GM
* possibly the internship timeline (though this is a bit more complicated)
* Tech Gala
* ...

mattxwang:
@advaithg @annaguo1012 what do y'all think?

mattxwang:
> just to clarify, this would have
> * a prop for the time at which the countdown ends
> * a child component prop for what to render before the time
> * another child component for what to render after the time

Not exactly - since I'm thinking of rendering the children in the "after" time, you wouldn't be able to put the "before" time content in there. I'm thinking instead of just having it as a component prop:

```jsx
<TimeRenderer time={123} fallback={<span>come back soon :))</span>}>
  <p>thanks for waiting</p>
</TimeRenderer>
```

Open to suggestions though!

> additionally, would this be displaying a timer of some sort for the time left as a countdown? (like the one on the GM page)
perhaps a prop to decide if a timer should be visible or not would also be a good idea - since for some cases we may not want a countdown timer (especially if we use it for the internship page in the future)

The timer could be passed to the prop for what to render before the time. We could separately create a reusable timer; I think a separation of concerns here is a good idea (especially since, as you point, out, we may not want a countdown timer).

advaithg:
sounds good! sounds like the tougher part of this ticket will be the time logic then

mattxwang:
@reginawang3495 two quick questions:

1. Is this ~ the level of difficulty that you want?
2. Is this description enough / too little / too much?

reginawang3495:
Looks good! A tiny bit wordy but is good difficulty level :)

reginawang3495:
And thanks for making tickets!

PR #365, opened 2021-11-30
Add countdown element

rmboyce:
Add countdown element to address issue #339. Also added the component to the old town hall page to test it. Takes in a prop for the time when the countdown ends as a Javascript timestamp (number of milliseconds from 1 January 1970 00:00:00 UTC, what you get from using valueOf for a Date), and a fallback component to render. When the time passes, the element switches from rendering the fallback component to its children components. Using the Javascript timestamp makes it resistant to timezone issues.

### DIFF
--- a/components/CountdownElement.js
+++ b/components/CountdownElement.js
@@ -3,8 +3,8 @@ import Countdown from 'react-countdown';
 
 export default function CountdownElement(props) {
     const chooseWhatToRender = ({ completed }) => {
-        if (completed) { 
-            return <>{props.children}</>; 
+        if (completed) {
+            return <>{props.children}</>;
         }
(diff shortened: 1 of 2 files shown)


advaithg (changes requested):
Sorry for the delayed review!

The countdown element seems to be working perfectly - my only concern is that the Date format could use some clarification. I tested it using a separate component as the prop, and with different dates and times.

Do you think you could add a comment explaining the format? (do this in the component itself)
Once we do that, we should be good to merge!

rmboyce:
I added a comment explaining the format.

advaithg (approved):
Looks good to me! Feel free to squash and merge.

PR #378, opened 2021-12-31
Added Search and Filter Functionality to Events Page

### LINKED ISSUE #305: Implement filter and search options for the Events Page
An end-user should be able to search and filter the events page to get a certain subset of events; I'm thinking:

* filter by committee -> the most important
* filter by time
* filter by location (in-person/virtual)

In addition, a naive text search (maybe title, location, committee, description) would be great!

Ideally, implementing the "backend" of this isn't too tricky; we can just apply a `.filter()` to the `events` array that's passed into the calendar. The trickier portions are going to be:

* developing the UI for this
* figuring out what to pass to the `filter` function - you'll probably want to do some sort of predicate-like writing!

The two most relevant files for you will be:

* [`pages/events.js`](https://github.com/uclaacm/website/blob/main/pages/events.js), the React component that manages events / holds the events page
* [`data/events.js`](https://github.com/uclaacm/website/blob/main/data/event.js), to peep the data structure of events right now

If you're curious, there's a more in-depth explanation of the data structure in #281.

## Stretch Goal

There's also a stretch goal; if you'd like to have a better search view than what we have right now, you could look at Angela's amazing mockup:

Screen Shot 2021-10-04 at 1 59 36 PM

Let me know if you need access to the Figma for this; we can add this as a "view" to `react-big-calendar`, and generate the cards relatively easily off of event data.

nendow02:
Closes #305.


nendow02:
Ready for review now

reginawang3495:
will review sometime tom hopefully !

components/Events/Filters.js
```diff
@@ -0,0 +1,142 @@
...
+    const handleLocationChange = (loc) => {
+        const updated = {...chosenLoc};
+        updated[loc] = !updated[loc];
+        setChosenLoc(updated);
+    };
+
+    return (
+        <div className={styles.filter}>
+            <div className={styles.inner}>
+                <div>
```
Nhamakami7, line 87: for code cleanliness, could prob get away with less <div>'s

components/Events/Filters.js
```diff
@@ -0,0 +1,142 @@
+import {faChevronUp, faChevronDown} from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useState, React } from 'react';
+import events from '../../data/event';
+import styles from '../../styles/components/Events/Filters.module.scss';
+
+
+export default function filters(props) {
+	const days = ['Sunday','Monday','Tuesday','Wednesday','Thursday','Friday','Saturday'];
```
reginawang3495, line 9: Can we do the same as L13? Basically same idea + code consistency + easier to read through code / debug if we can print out `chosenDays`

components/Events/Filters.js
```diff
@@ -0,0 +1,142 @@
...
+	const [filterOpen, setFilterOpen] = useState(false);
```
reginawang3495, line 10: nit: prefer to start boolean values with `is`, so we can rename this to `isFilterOpen`

components/Events/Filters.js
```diff
@@ -0,0 +1,142 @@
...
+            if (String(item.committee).toLowerCase().includes(lowerCaseSearch)) return true;
+            if (String(item.location).toLowerCase().includes(lowerCaseSearch)) return true;
+            if (String(item.description).toLowerCase().includes(lowerCaseSearch)) return true;
+            return false;
+        });
+        //Check filters
+        const timeEmpty = fromTime === '' || toTime === '';
+        const commEmpty = Object.keys(chosenComms).every(comm => !chosenComms[comm]);
+        const daysEmpty = chosenDays.every(day => !day);
+        const locEmpty =  Object.keys(chosenLoc).every(loc => !chosenLoc[loc]);
```
reginawang3495, line 46: nit: prefer to start boolean values with `is`

Took me a bit to realize what these variables meant

components/Events/Filters.js
```diff
@@ -0,0 +1,142 @@
...
+                                <input type='time' className={styles.timeinput} value={fromTime} onChange={(e) => setFromTime(e.target.value)} />
+                                <br/>
+                                <input type='time' className={styles.timeinput} value={toTime} onChange={(e) => setToTime(e.target.value)} />
+                            </div>
+                        </div>
+                    </div>
+                    <div>
+                        <h3 className={styles.header}>Location</h3>
+                        <input type="checkbox" id="online" checked={chosenLoc.online} onChange={() => handleLocationChange('online')}/> Online <br/>
+                        <input type="checkbox" id="in_person" checked={chosenLoc.in_person} onChange={() => handleLocationChange('in_person')}/> In-Person
```
reginawang3495, line 136: Can use `Object.map` so that if, for instance, someone modifies the L135, they won't forget to modify the other line (since they will be collapsed together)

components/Events/Filters.js
```diff
@@ -0,0 +1,142 @@
...
+                        <h3 className={styles.header}>Time</h3>
+                        <div className={styles.timecontainer}>
+                            <div>
+                                <div>From </div>
+                                <div>To </div>
+                            </div>
+                            <div>
+                                <input type='time' className={styles.timeinput} value={fromTime} onChange={(e) => setFromTime(e.target.value)} />
+                                <br/>
+                                <input type='time' className={styles.timeinput} value={toTime} onChange={(e) => setToTime(e.target.value)} />
```
reginawang3495, line 129: Consider collapsing these two

reginawang3495 (changes requested):
looks really great!! thanks for working on this nathan :))

here are a few changes to make the code easier to read !

components/Events/Filters.js
```diff
@@ -0,0 +1,204 @@
...
+
+export default function filters(props) {
+  const [isFilterOpen, setIsFilterOpen] = useState(false);
+  const [searchValue, setSearchValue] = useState('');
+  const [chosenComms, setChosenComms] = useState({
+    studio: false,
+    icpc: false,
+    design: false,
+    cyber: false,
+    teachLA: false,
```
reginawang3495, line 15: We can do this to prevent having to do a workaround like L75-L76 of this file
```suggestion
    "teach-la": false,
```

components/Events/Filters.js
```diff
@@ -0,0 +1,204 @@
...
+    Wednesday: false,
+    Thursday: false,
+    Friday: false,
+    Saturday: false,
+  });
+  const [fromTime, setFromTime] = useState('');
+  const [toTime, setToTime] = useState('');
+  const [chosenLoc, setChosenLoc] = useState({
+    Online: false,
+    In_person: false,
```
reginawang3495, line 34: Same with this
```suggestion
    "In-Person": false,
```

components/Events/Filters.js
```diff
@@ -0,0 +1,204 @@
...
+  });
+
+  const handleSearch = () => {
+    let allEvents = [...events];
+    //Check search input
```
reginawang3495, line 39: nit: generally, people often have a space between `//` and the comment itself

components/Events/Filters.js
```diff
@@ -0,0 +1,204 @@
...
+      if (String(item.title).toLowerCase().includes(lowerCaseSearch))
+        return true;
+      if (String(item.committee).toLowerCase().includes(lowerCaseSearch))
+        return true;
+      if (String(item.location).toLowerCase().includes(lowerCaseSearch))
+        return true;
+      if (String(item.description).toLowerCase().includes(lowerCaseSearch))
+        return true;
+      return false;
+    });
```
reginawang3495, line 51: Nice job on all the functional methods!! 

We can do something like ;)
```suggestion
    const lowerCaseSearch = searchValue.toLowerCase();
    let allEvents = events.filter((item) =>
      [item.title, item.committee, item.location, item.description].some(
        (itemsToSearch) =>
          String(itemsToSearch).toLowerCase().includes(lowerCaseSearch)
      )
    );
```

components/Events/Filters.js
```diff
@@ -0,0 +1,204 @@
...
+        isTimeEmpty || (itemMins >= fromMins && itemMins <= toMins);
+      const dayValid = isDaysEmpty || chosenDays[itemDay];
+      const commValid = isCommEmpty || chosenComms[itemComm];
+      const onlineValid = isLocEmpty || chosenLoc[itemLoc];
+      return timeValid && dayValid && commValid && onlineValid;
+    });
+    const newEvents = allEvents.map((original_event, index) => ({
+      ...original_event,
+      id: index,
+    }));
```
reginawang3495, line 88: Could we separate out each filter so that it's easy for people to add filters in the future if they want + for code readability? EG (I didn't test so it might not be right)
```
    // Check time filter
    const isFilterTimeEmpty = Object.keys(chosenTimes).every(
      (time) => time === ""
    );
    if (!isFilterTimeEmpty) {
      filterFromMins =
        parseInt(fromTime.split(":")[0]) * 60 +
        parseInt(fromTime.split(":")[1]);
      filterToMins =
        parseInt(toTime.split(":")[0]) * 60 + parseInt(toTime.split(":")[1]);

      allEvents = allEvents.filter((item) => {
        const itemStartDate = new Date(item.start);
        const itemStartMins =
          itemStartDate.getHours() * 60 + itemStartDate.getMinutes();

        const itemEndDate = new Date(item.end);
        const itemEndMins =
          itemEndDate.getHours() * 60 + itemEndDate.getMinutes();

        const timeValid =
          isTimeEmpty ||
          (itemStartMins >= filterFromMins && itemEndMins <= filterToMins);
        return timeValid;
      });
    }
```

components/Events/Filters.js
```diff
@@ -0,0 +1,204 @@
...
+                  <div key={loc}>
+                    <input
+                      type="checkbox"
+                      id="online"
+                      checked={chosenLoc[loc]}
+                      onChange={() => handleLocationChange(loc)}
+                    />{title}
+                  </div>
+                );
+              })}
```
reginawang3495, line 197: This can be simplified if we change to the `In-Person` key + can we make the `id` equal to `loc` or something?

reginawang3495 (changes requested):
good job! some thoughts for code cleanliness / readability

components/Events/Filters.js
reginawang3495: I also wouldn't mind putting
```
        const itemEndDate = new Date(item.end);
        const itemEndMins =
          itemEndDate.getHours() * 60 + itemEndDate.getMinutes();
```
in a separate method like `dateToMinutes` or something like that

components/Events/Filters.js
```diff
@@ -0,0 +1,228 @@
...
+      ...original_event,
+      id: index,
+    }));
+    props.handleChange(newEvents);
+  };
+
+  const handleWeekdayChange = (day) => {
+    const updated = { ...chosenDays };
+    updated[day] = !updated[day];
+    setChosenDays(updated);
```
reginawang3495, line 113: you can do something like this! can apply to other parts also
```suggestion
    setChosenDays({ ...chosenDays, [day]: !chosenDays[day] });
```

components/Events/Filters.js
```diff
@@ -0,0 +1,228 @@
...
+            </div>
+            <div>
+              <h3 className={styles.header}>Days</h3>
+              {Object.keys(chosenDays).map((day) => (
+                <div key={day}>
+                  <input
+                    type="checkbox"
+                    id={day}
+                    checked={chosenDays[day]}
+                    onChange={() => handleWeekdayChange(day)}
```
reginawang3495, line 181: There can potentially be a mismatch, so maybe taking the direct `checked` value would be more consistent! Same with the other checkboxes

```suggestion
                    onChange={(e) => handleWeekdayChange(day, e.target.checked)}
```

components/Events/Filters.js
```diff
@@ -0,0 +1,228 @@
...
+                return (
+                  <div key={loc}>
+                    <input
+                      type="checkbox"
+                      checked={chosenLoc[loc]}
+                      onChange={() => handleLocationChange(loc)}
+                    /> {loc}
+                  </div>
+                );
+              })}
```
reginawang3495, line 221: nit: can remove `return` for less code

components/Events/Filters.js
```diff
@@ -0,0 +1,228 @@
...
+  const handleSearch = () => {
+    // Check search input
+    const lowerCaseSearch = searchValue.toLowerCase();
+    let allEvents = events.filter((item) =>
+      [item.title, item.committee, item.location, item.description].some(
+        (itemsToSearch) =>
+          String(itemsToSearch).toLowerCase().includes(lowerCaseSearch),
+      ),
+    );
+    // Check time filter
```
reginawang3495, line 55: nit: spacing between chunks for readability
```suggestion

    // Check time filter
```

reginawang3495 (commented):
looks a lot neater! this is probably just some final thoughts

components/Events/Filters.js
```diff
@@ -0,0 +1,222 @@
...
+    Monday: false,
+    Tuesday: false,
+    Wednesday: false,
+    Thursday: false,
+    Friday: false,
+    Saturday: false,
+  });
+  const [chosenTimes, setChosenTimes] = useState({
+    fromTime: '',
+    toTime: '',
```
reginawang3495, line 32: nit, no need to change if you don't want to: 
```suggestion
    from: '',
    to: '',
```
Since `Time` is redundant. It's not a big deal so I don't mind leaving as is

components/Events/Filters.js
```diff
@@ -0,0 +1,222 @@
...
+    const isFilterTimeEmpty = Object.keys(chosenTimes).every(
+      (time) => chosenTimes[time] === '',
+    );
+    if (!isFilterTimeEmpty) {
+      const filterFromMins =
+        parseInt(chosenTimes.fromTime.split(':')[0]) * 60 +
+        parseInt(chosenTimes.fromTime.split(':')[1]);
+      const filterToMins =
+        parseInt(chosenTimes.toTime.split(':')[0]) * 60 +
+        parseInt(chosenTimes.toTime.split(':')[1]);
```
reginawang3495, line 65: I think there's a possibility that `from` is set, but not `to` and I don't think this covers that edge case?

styles/components/Events/Filters.module.scss
```diff
@@ -0,0 +1,72 @@
+@import '../../variables.scss';
```
reginawang3495, line 1: @advaithg I don't think this look as you might want on phone. Not sure how much you care about that / if you wanted to review because of that!
CAD9037E-442D-4109-9D98-3BDF792CACEC

reginawang3495 (changes requested):
@advaithg did you have any thoughts as well?

styles/components/Events/Filters.module.scss
advaithg: Yeah I think this is not a great idea to have on mobile, so I see two options here:

1. Hide the filter option entirely on mobile
2. Fix it so search and filter buttons are underneath the search bar + centered

either way, I'd rather have a feature non-existent on mobile than not responsive
I think I'd prefer option 2, but if you would like to get this PR merged in, you can also do 1 and open a separate issue for 2 later!

advaithg (commented):
Nit and not necessarily required, but I think it would be nice for the filter menu to close itself once a user hits search (if I'm not wrong, that's how it usually works)

styles/components/Events/Filters.module.scss
```diff
@@ -0,0 +1,91 @@
...
+  background-color: white;
+  border: 1px solid lightgray;
+  border-radius: 0 0 1rem 1rem;
+  display: grid;
+  grid-template-columns: 25% 25% 25% 25%;
+  justify-content: space-evenly;
+  padding: 0 1rem 1rem;
+  position: relative;
+  @media screen and (max-width:1024px) {
+    grid-template-columns: 40% 40%;
```
advaithg, line 33:
```suggestion
    grid-template-columns: 40% 40%;
    text-align: left;
```

styles/components/Events/Filters.module.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+  @media screen and (max-width: 434px) {
+    grid-template-columns: 100%;
```
advaithg, line 36:
```suggestion
    grid-template-columns: 100%;
    text-align: left;
```

styles/components/Events/Filters.module.scss
```diff
@@ -0,0 +1,91 @@
...
+  right: 0;
+}
+
+.filteroptions {
+  background-color: white;
+  border: 1px solid lightgray;
+  border-radius: 0 0 1rem 1rem;
+  display: grid;
+  grid-template-columns: 25% 25% 25% 25%;
+  justify-content: space-evenly;
```
advaithg, line 29:
```suggestion
  justify-content: space-evenly;
  margin-top: 10px;
```

mattxwang (commented):
I'm not part of the review chain (so feel free to ignore me), but two general thoughts on my mind:

* can we reduce the filter logic to several composed filters (ex one for committee, one for chosen day, etc.)? I would argue that it's easier to read and test, but maybe it feels unnatural. Also particularly good if we add unit tests in the future.
* can we use a reducer and/or programmatically generate the state more *from* the data? Right now I see a lot of object hard-coded state, and I'm not sure if that's the best approach.

reginawang3495:
@nendow02 anything I can do to support this? it's almost done and it'd be cool to merge in!

nendow02:
> @nendow02 anything I can do to support this? it's almost done and it'd be cool to merge in!

I’m almost done with the new requested changes. One thing I may have issues with though is using a reducer since I haven’t really implemented that before so I might need help on that.

components/Events/Filters.js
```diff
@@ -0,0 +1,236 @@
...
+  };
+  const [filterVals, dispatch] = useReducer(reducer, initialState);
+  const [isFilterOpen, setIsFilterOpen] = useState(false);
+  const [searchValue, setSearchValue] = useState('');
+
+  const dateToMinutes = (date) => {
+    const itemMins =
+      date.getHours() * 60 + date.getMinutes();
+    return itemMins;
+  };
```
reginawang3495, line 60: I'd put this outside of `export default function filters(props) {` since it's not specific to this component and can theoretically be used elsewhere

components/Events/Filters.js
```diff
@@ -0,0 +1,236 @@
...
+
+      return filterEvents.filter((item) => {
+        const itemStartMins = dateToMinutes(new Date(item.start));
+        const itemEndMins = dateToMinutes(new Date(item.end));
+        return itemStartMins >= filterFromMins && itemEndMins <= filterToMins;
+      });
+    }
+    return filterEvents;
+  };
+  const filterComm = (filterEvents) => {
```
reginawang3495, line 83: nit: use new lines to cleanly separate and make code easier to read
```suggestion

  const filterComm = (filterEvents) => {
```

components/Events/Filters.js
```diff
@@ -0,0 +1,236 @@
+import { faChevronUp, faChevronDown } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useState, useReducer, React } from 'react';
+import events from '../../data/event';
```
reginawang3495, line 4: nit: use `allEvents` instead to avoid confusion with using `events` in other places
```suggestion
import allEvents from '../../data/event';
```

components/Events/Filters.js
```diff
@@ -0,0 +1,236 @@
...
+  const [filterVals, dispatch] = useReducer(reducer, initialState);
+  const [isFilterOpen, setIsFilterOpen] = useState(false);
+  const [searchValue, setSearchValue] = useState('');
+
+  const dateToMinutes = (date) => {
+    const itemMins =
+      date.getHours() * 60 + date.getMinutes();
+    return itemMins;
+  };
+  const filterTime = (filterEvents) => {
```
reginawang3495, line 61: `filterEvents` is a bit confusing, it sounds like filter is an event. I think `eventsToFilter` also works, but it's a bit long
```suggestion
  const filterTime = (events) => {
```

components/Events/Filters.js
```diff
@@ -0,0 +1,236 @@
...
+  const handleSearch = () => {
+    // Check search input
+    const lowerCaseSearch = searchValue.toLowerCase();
+    let allEvents = events.filter((item) =>
+      [item.title, item.committee, item.location, item.description].some(
+        (itemsToSearch) =>
+          String(itemsToSearch).toLowerCase().includes(lowerCaseSearch),
+      ),
+    );
+    allEvents = filterTime(allEvents);
```
reginawang3495, line 127: it isn't `allEvents` in that it's not all possible events from the start, probably rename

pages/events.js
```diff
@@ -26,11 +27,12 @@ const getEventClassByEvent = (event) => {
 	});
 };
 
-const indexedEvents = events.map((original_event, index) => ({...original_event, id: index}));
 const googleCalendarShare = 'https://calendar.google.com/calendar/u/2?cid=YWNtYnJ1aW5zQGdtYWlsLmNvbQ';
 
 function Events() {
 	const [activeEvent, setActiveEvent] = useState(null);
+	const [indexedEvents,setIndexedEvents] = useState(events.map((event, index) => ({...event, id: index})));
```
reginawang3495, line 34:
```suggestion
	const [indexedEvents, setIndexedEvents] = useState(events.map((event, index) => ({...event, id: index})));
```

reginawang3495 (changes requested):
lgtm as long as we revert `yarn.lock` changes! Everything else looks great

reginawang3495 (approved):
I think you deleted the yarn.lock :0 if you could put that back !

looks good to me, feel free to merge!! Thank you for sticking to it—i know there’s been a ton of back and forth which can get p tiring :)

reginawang3495:
mb looks good to me! pls feel free to merge, I think I accidentally selected commit incorrectly

PR #385, opened 2022-01-07
change banner, say GM is done

advaithg:


pages/index.js
```diff
@@ -32,7 +32,7 @@ function Home () {
 			/>
 			<div className="home-page text-center">
 				<ContentBanner
-					title="ACM's Winter General Meeting is coming soon!"
+					title="Check out our Winter General Meeting!"
```
evanzhong, line 35: Might want to reword?

`See what happened?` something more indicative of the past?

evanzhong (approved):
lgtm, minor comment

pages/index.js
advaithg: Oh yeah that's a good idea

Issue #371, opened 2021-12-15
Remove image dependency on Teach LA website, switch to Image tag


mattxwang:
When I wrote the hack-y implementations for the JEDI and Impact pages, I used people's headshots from the Teach LA website when possible. While it made prototyping for me easier, this is clearly a bad thing to do (since we don't control that codebase).

So, we should move those images to our own site. On top of that, this could give us an opportunity to switch the Impact and JEDI officer cards to use the `<Image />` tag, now that we'll know what the images look like at site compile time.

## Relevant Files

Impact officers:

https://github.com/uclaacm/website/blob/fba4833bce24fd70c5e007a06f1984251f95f10e/data/impact.js#L29-L84

JEDI officers:

https://github.com/uclaacm/website/blob/fba4833bce24fd70c5e007a06f1984251f95f10e/data/jedi.js#L126-L162

advaithg:
JEDI page images addressed with #392

PR #394, opened 2022-01-13
Issue 371: Moved Teach LA images to our own site and used the Image tag

ichettiath:
The profile pics of the Teach LA officers were taken from Teach LA's website. The images were moved to our own site and `` tags were replaced with `<Image />` tags.


advaithg (changes requested):
This looks good! Thanks for doing this + also switching to `Next/Image`
The only change I have is to reorder the `Image from next\image` import to be above the react import, like the warning says.

advaithg (commented):
something like "closes #371" or "resolves #371" in the issue description will automatically close the issue as well

components/Impact/PersonCard.js
```diff
@@ -6,7 +7,7 @@ export default function PersonCard({name, pronouns, position, major, year, img,
 		<div className={`${styles['workshop-card']} ${styles['grid-tablet-only-2']}`}>
 			<div className={styles['workshop-image-container']}>
 				{/* eslint-disable-next-line @next/next/no-img-element */}
```
mattxwang, line 9: Should remove this comment now

components/JEDI/JEDICard.js
```diff
@@ -11,7 +12,9 @@ function JEDICard({ name, pronouns, bio, pic, alt }) {
   return (
     <div className={styles['mb-2']}>
       {/* eslint-disable-next-line @next/next/no-img-element */}
```
mattxwang, line 14: Should also remove this comment now

mattxwang (commented):
I think there's some cleanup for this PR that needs to be done, @advaithg @ichettiath

PR #408, opened 2022-01-25
Fixes to event generation script

advaithg:
As of now
- changes location to zoom and adds zoom link if event is occuring on zoom (checks for `ucla.zoom.us`)
- removes first line of CSV file so you don't have to manually remove when saving
- removes whitespaces before and after committee names when mapping

Going to leave a lot of other problems with this page to try and deal with them during the one click event publishing project, especially the ones regarding hardcoding - since I'd like to include the events page to be one of the targets of publishing!

(sorry for the unnecessary commits, kinda confused myself on the links logic)

### DIFF
--- a/scripts/lib.mjs
+++ b/scripts/lib.mjs
@@ -51,9 +51,17 @@ const generateSingleEvent = ({
     end = moment(`${date} ${endHr} PM`, 'YYYY-MM-DD LT').valueOf();
   }
 
-  // add if location contains ucla.zoom.us, change location to "zoom" and links to have the zoom link
-  if (!links) {
-    links = [];
+  if(location.includes('ucla.zoom.us')) {
+    let zoomLink = location;
+    location = 'Zoom';
+    if(!links){
+      links = [];
+    }
+    links.push({
+      text: 'Zoom Link',
+      href: zoomLink,
+      ext: true,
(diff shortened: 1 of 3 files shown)


scripts/lib.mjs
```diff
@@ -51,16 +51,24 @@ const generateSingleEvent = ({
...
-    if (fblink) {
-      links.push({
-        text: 'Facebook Event',
-        href: fblink,
-        ext: true,
-      });
-    }
+  }
+  if(location.includes('ucla.zoom.us')) {
+    let zoomLink = location;
```
mattxwang, line 58: nit:
```suggestion
    const zoomLink = location;
```

mattxwang (commented):
Seems reasonable to me, small nit but other than that should be good to merge.

Have you tried this on last week's events, and testing the outputs side-by-side?

advaithg:
> Have you tried this on last week's events, and testing the outputs side-by-side?

Yep! Works fine except for the impact fireside chats `tinyurl` edge case, which it isn't supposed to work on anyway.

mattxwang (approved):
Good work!

Eventually we should have test cases for this. Maybe something we can follow up on after I start merging in some of my accessibility testing work.

PR #423, opened 2022-02-09
Update carousel etc

MiraynaMckinney:
Start Updating Carousel Images


reginawang3495 (approved):
lgtm!!

@advaithg I wonder if we should have data on alts for each image (maybe a separate ticket)

advaithg:
> @advaithg I wonder if we should have data on alts for each image (maybe a separate ticket)

I don't think so, since 
- these images are purely decorative
- they are moving, so alt text might be needlessly confusing

I think having `alt=""` to mark as decorative works.

advaithg (commented):
hmm it looks like `about2.png` will probably have to be manually rotated or angled (as `about1.png` was) so it looks upright in the website

advaithg (commented):
Hi @MiraynaMckinney , just checking to see if you're still working on this PR!

PR #443, opened 2022-02-28
adds explore moonshot event

advaithg:


mattxwang (approved):
Thanks for the quick turnaround on this!

PR #452, opened 2022-03-14
Updates new leadership!

advaithg:


data/officers.js
```diff
@@ -1,72 +1,87 @@
...
+		name: 'Milo Kearney',
 		title: 'President, Teach LA',
-		image: '/images/officers/SophieSchoenmeyer.jpg',
+		image: '/images/officers/MiloKearney.jpg',
+	},
+	{
+		name: 'Eden Yu',
+		title: 'Vice President, Teach LA',
+		image: '/images/officers/EdenYu.jpg',
+	},
```
mattxwang, line 76: Typically, we don't include VPs on the leadership page - though feel free to override me on this decision! It does seem that the other committees with VPs (Design, ACM-W) swapped over to doing Co's.

mattxwang (changes requested):
Looks like this is WIP, but just do jot some things down (+ the 1 review comment):

* you'll want to make sure that all the images are square; if they aren't, you'll get squishing. Amy's is the most obvious example, but it seems like many of the photos aren't square as of now
* you probably also want to compress and downsize the images for what they're actually needed for. this iteration of the page is pretty small avatar icons, so you definitely don't need anything > 512px; ideally, none of the images should be over a MB
* there are missing images, but I imagine you're already aware of that

You can pass this off to Matt, but there are ways that you can enforce the above - I wrote an issue for this in the Teach LA website (https://github.com/uclaacm/teach-la-website/issues/230) that can be a good starting point.

data/officers.js
matthewcn56: On the topic of adding people to the leadership page, I know that Teach LA has a new role split of Milo as the TLA General President and Jiin as the TLA Dev President, so I think that there's a case to be made for her to be added onto the leadership page as well!

advaithg:
>  you'll want to make sure that all the images are square; 

already reached out to Ava to ask the new cprez for better images in these cases!

>  you probably also want to compress and downsize the images for what they're actually needed for. 

will do!

>   there are missing images, but I imagine you're already aware of that

yep!

> You can pass this off to Matt

@matthewcn56 you down to finish up this PR? I can push the few more updated images I got!

data/officers.js
advaithg: I decided on adding the TLA VP since none of the other committees as of now have VPs iirc, but I will reach out to Ava to confirm! 
Also think it's a good idea to add TLA Dev!

matthewcn56:
> @matthewcn56 you down to finish up this PR? I can push the few more updated images I got!

Sure! I can address the changes requested once you add in the updated images (and once finals ramp down for me after tmrw haha)

data/dev.js
```diff
@@ -1,31 +1,34 @@
+/* please have the dev team director as the first element of this array, so that the 
+dev team page correctly autopopulates the director's data! */
 export const people = [
-  {
-		name: 'Advaith Gowrishetty',
+//dev team director also has an extra email field as they are the point of contact for the dev team
```
mattxwang, line 4: Do you want to encode this somewhere else instead of in this field? Ex a constant somewhere? I feel like this complicates the `people` design.

advaithg: Also agree with this ^ would be easier to just have a constant in the page

matthewcn56: Sure, this can be set as a constant within the `dev.js` page instead

matthewcn56:
In addition to compressing the new images, making them square, and updating them for the new year, the `officers` map uses `officer.image` instead of `officer.email` as the email field is not presently filled for all of our officers. 

Also slightly changed the `dev.js` page to turn the dev team director into a constant object that can be easily updated.

mattxwang:
Did you downsize the image as well? I see some are > 2k by 2k, I'm not sure if they need to be that big.

You can batch resize images with Photoshop or something like optipng/jpegoptim, don't do it by hand haha.

matthewcn56:
> You can batch resize images with Photoshop or something like optipng/jpegoptim, don't do it by hand haha.

I ran it through optipng /jpegoptim, but it just decreased the sizes of the image files, not the dimensions themselves

mattxwang:
> I ran it through optipng /jpegoptim, but it just decreased the sizes of the image files, not the dimensions themselves

Oh that's unfortunate, do you have Photoshop installed? That's how I did it for Teach LA a while ago!

mattxwang (approved):
There are some minor size mismatches, but I don't think it's too big of a deal. LGTM - thanks for your work on wrapping this up @matthewcn56!

mattxwang:
Hey, just realizing - [this broke the impact page](https://www.uclaacm.com/impact) - see how Alyssa's image is now a 404 since the file extension changed. 

cc @matthewcn56

PR #454, opened 2022-03-23
changes events page `.filter` z-index to 9

mattxwang:
This is a bugfix for something that has been bothering me: the search bar/filter dropdown on the events page is "on top" of the navbar:

Screen Shot 2022-03-23 at 1 04 27 PM
Screen Shot 2022-03-23 at 1 04 37 PM

This PR lowers the z-index on `.filter` to `9`, so it's less than the navbar (which is `10`).

Screen Shot 2022-03-23 at 1 05 19 PM

### DIFF
--- a/styles/components/Events/Filters.module.scss
+++ b/styles/components/Events/Filters.module.scss
@@ -3,7 +3,7 @@
 .filter {
   padding-bottom: 2.5rem;
   position: relative;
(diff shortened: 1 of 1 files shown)


nendow02 (approved):
Looks good to me, thanks for the fix.

PR #467, opened 2022-04-05
change banner to point to officer apps

advaithg:
Updates website to reflect officer applications for S22!

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -32,9 +32,9 @@ function Home () {
 			/>
 			<div className="home-page text-center">
 				<ContentBanner
-					title="See what happened at our Winter General Meeting!"
-					ctaText="Learn More"
-					ctaLink="/gm/w22"
+					title="ACM Spring Officer Applications are out!"
+					ctaText="Apply"
+					ctaLink="https://tinyurl.com/acmspringofficer"
 				/>
 				<Banner />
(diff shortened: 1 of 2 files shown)


mattxwang:
Do you want to change the "join us" page as well?

matthewcn56:
> Do you want to change the "join us" page as well?

Maybe instead of changing the join us page, we can just change the link that the `join us` points to within the navbar instead?

advaithg (commented):
Oh yeah I will also update the internship page to mention the officer apps, since we have a place for that on that page too.

advaithg:
> Maybe instead of changing the join us page, we can just change the link that the `join us` points to within the navbar instead?

will check with Ava on this! @avaasmani

matthewcn56:
@advaithg on the join us page, maybe the headline on the top that says that some committees are recruiting can contain a link to the Apply Now card lower down in the page or directly to the application? As is, it can be a bit hard for people to find where the application is without scrolling down.

matthewcn56 (approved):
Lgtm, thanks for adjusting the info to point to our officer apps! Up to you if you want to wait for ava to take a looksie at it too before merging in.

avaasmani:
Yeah we will keep the join us page pointing to the internship page, the banner looks good!

PR #469, opened 2022-04-06
fix sidebar and change year in footer.

dtjanaka:


dtjanaka:
Interviewees pointed this out during dev team interviews.

matthewcn56 (approved):
Thanks for fixing this Dylon! I think it's great how most things got moved to more responsive sizing, and fixing our sidebar. I think the media query to turn off the sidebar is pretty clever and thanks for fixing the year :))

mattxwang:
Just to chime in - from my understanding, this isn't really how copyright works. You should be specifying a start and end date, or just the start - by just putting 2022, you're actually forgoing your claim to copyright from the year before (though IANAL, and this has never been litigated in the US).

Perhaps what we should do is find when our chapter was established, and then put a range from then until the current year?

matthewcn56:
> Perhaps what we should do is find when our chapter was established, and then put a range from then until the current year?

Ooh, thanks for letting us know Matt! I'll write out an issue to fix that as well, i did not know that's how copyright should work!

dtjanaka:
Interesting. At any rate, the newsletter also has an incorrect year.
C42E6DE6-2EB1-4531-B4E0-A9FE7371B9A7

PR #471, opened 2022-04-07
changes navbar wordmark to include 'chapter'

mattxwang:
As requested by the ACM. This is a pretty rudimentary job; I think if we want to keep this long-term, we should ask design to make us a wordmark with the proper kerning and font-weight (there's no production-ready Westwood Sans, so I've had to use the regular font weight).


matthewcn56 (approved):
Thanks for addressing this! I agree, we should def ask design to make us a proper wordmark in the future but this works in the short term given what higher ACM is saying.

PR #473, opened 2022-04-08
ACM at UCLA --> ACM chapter at UCLA on everything else 

dtjanaka:
exceptions: alt text and old events


matthewcn56 (changes requested):
There's just one more place that i think needs to be changed within the README on line 142!

mattxwang:
Wait sorry, can we ... *not* do this yet? I'm looking to meet w/ the ACM to renegotiate the branding guidelines, and as of now our website is compliant with what they've asked.

PR #502, opened 2022-05-08
Fixed the font-size in the About page.

0Kwanele0:
Paragraphs font-size, in the about page, was small (0.8em) yet it was suppose to be 1em just like all the other pages.

### DIFF
--- a/styles/pages/About.module.scss
+++ b/styles/pages/About.module.scss
@@ -86,7 +86,7 @@
   }
 
   p {
-    font-size: 0.8em;
+    font-size: 1em;
     line-height: 150%;
     margin: 0 22px 14px;
   }


mattxwang (changes requested):
Definitely want to defer to @matthewcn56 / @uclaacm/website-team, but I do think this is a great idea! However, it may require a bit of tweaking - this new change creates a horizontal scrollbar and/or poorly overflows on viewports in the range ~ `[650px, 950px]`. See below (note the diamond not extending across the page - it has `width: 100%`):

Screen Shot 2022-05-13 at 3 30 11 PM


Screen Shot 2022-05-13 at 3 30 21 PM
 

My suggestion: let's loosen the restricted `max-width` of the about page, and remove all absolute sizing references - this problem *also* exists on live if someone changes the base font size for accessibility reasons (and thus adjusts the `em` unit definition without resizing the image).

PR #412, opened 2022-01-28
ci: minimum working example for jest-axe accessibility CI

mattxwang:
This PR implements the minimum overhead required to run `jest-axe` via GitHub Actions on a set of predefined components. In order, we:

1. install `jest`, `jest-axe`, `@testing-library/react`, `@testing-library/jest-dom`, the minimum set up for a viable test infrastructure
    1. to install `jest`, we add a general `jest.config.js` as a jest configuration file, and `jest.setup.js` as a pre-test setup file (automatically importing our dependencies)
    2. creates a set of mock module mappers for "non-code" webpack elements; a stub file mock (`__mocks__/fileMock.js`) for most files, and the `identity-obj-proxy` to proxy-back CSS/SASS/SCSS class names
    3. adds the `next` preset to babel
    4. adjusts the ESLint config to use the jest environment
2. implements a rudimentary accessibility test in `NewsArticle.test.js`, which is a simple subcomponent with no errors. You can validate this working by adding an arbitrary `` to `NewsArticle.js`; `npm test` will throw an error.
3. implements a rudimentary failing accessibility test in `Footer.test.js`, where an `<ul>` has children that are not `<li>` - showing that negatives work. I then resolve this issue.
4. updates the "Node.JS CI" action to be "Build and Test", and runs `npm test` in it

Ideally, you should be able to add more test files following the configuration in `NewsArticle.test.js`; note the strange `let` pattern with the async version of `act`.

I didn't use the native next testing plugin, since that seems blocked by the upgrade to Next 12 (#409).

Part of #218, supersedes #301.

Relevant reading:

* [Configuring Jest](https://jestjs.io/docs/configuration)
* [`act()` in React](https://reactjs.org/docs/test-utils.html#act)
* [Blog Post: Next.js 11: Setup & Config for Testing, Linting, and Absolute Imports](https://benjaminwfox.medium.com/next-js-setup-config-for-testing-linting-and-absolute-imports-605959d7bd6f)

PR #507, opened 2022-06-15
tests: adds aXe test for `<Officers />`; resolves `key` bug

mattxwang:
Followup to #412, gradually increasing coverage.

Two small notes:

1. You need to mock `next/image` somehow. This trivial mock is from https://github.com/vercel/next.js/discussions/18516, however I think after you upgrade to Next 12 it becomes easier.
2. There's a current bug in `Officers` that incorrectly uses a non-existent field as the `key`. Updated to use an existing field.


mattxwang:
Just going through my own stale PRs, bump @matthewcn56.

smalex-z (approved):
lgtm! full axe support not expected to be supported for some time, but this PR contributes.

PR #531, opened 2022-09-14
Fall'22 gm page created

jainsujay02:
- Added page for fall GM 2022
- Basic content updates and some alt text changes for improved accessibility.


jainsujay02:
I will update the website banner and footer once ASUCLA approves the marketing graphic.

matthewcn56 (changes requested):
Looks good, thanks for doing this! Just left the comment of changing the image tag but I think it should be fine after that

PR #550, opened 2022-11-11
Footer links updated

jainsujay02:
Footer links updated.


matthewcn56 (approved):
Thanks for removing some of the extraneous links, this is a good step in helping to reduce our footer size!

PR #552, opened 2022-11-11
about-page-updates

jainsujay02:
-Increased font size for the body text
- Added a What's ACM board? -Added a What are ACM Initiatives?
 -Add a quick explanation for Leadership.
-Added space before footer


matthewcn56 (approved):
Other from the lint warnings I think this looks fine, thanks for taking the time to update this Sujay!

PR #577, opened 2023-01-04
2023 Winter GM page added

jainsujay02:
-Added Winter 2023 GM page.


matthewcn56 (approved):
Looks good to me, thanks for doing this! There's just a couple of things such as some warnings that the linter is leaving about imports but they can be fine to leave in. I'm making a mental note that there can be some restructuring we can do like putting repeated logic across gm pages inside separate files, but that's something I'll make our interns do next quarter and can be moved to a different PR!

PR #579, opened 2023-01-04
Changes to leadership in about page

jainsujay02:


data/leadership.js
```diff
@@ -85,8 +80,8 @@ export default [
 		img: '/images/officers/MuthuPalaniappan.jpg',
 	},
 	{
-		name: 'Amy Seo',
+		name: 'Ivy Kang',
 		position: 'Co-President, Design',
-		img: '/images/officers/AmySeo.png',
+		img: '/images/officers/IvyKang.JPG',
```
matthewcn56, line 85: All of the images we put in are compressed to reduce file size, could you please run her picture  through a compressor like https://imagecompressor.com/ ?

matthewcn56: Adding on top of this, the way that we format images of officers, we ideally would like the images to be of square shape, can you please edit Ivy's picture so that it's of square size? Screenshot 2023-01-03 at 11 03 04 PM

matthewcn56 (changes requested):
Thanks for the udpates! Just left in some comments about compressing Ivy's pic to reduce our page size and making it a square so that it doesn't get warped.

data/leadership.js
jainsujay02: Thank you for catching this!

matthewcn56 (approved):
Great, thanks for making these changes!

PR #563, opened 2022-11-24
Add script that uses google sheets api to read newsletter events

Amyh11325:
# Overview
- `event-generator-sheets.mjs` is a script similar to `recurring-event-generator.mjs` and `single-event-generator.mjs` but instead of reading data from a .cvs file downloaded from the ACM Newsletter spreadsheet, this script uses Google's API to read data directly from the spreadsheet.

# Testing
Run `node scripts/event-generator-sheets.mjs`
Check the output written to `output.json`
```
[{"title":"Studio Fall General Meeting","allDay":false,"start":1664503200000,"end":1664508600000,"location":"Haines Hall A18","committee":"studio","description":"Do you enjoy playing games? Want to learn how to make them? 🎮 💖 ACM Studio is hosting our Fall General Meeting this Thursday, September 29th @7:00PM in Haines A18. Here you can meet our wonderful board of officers and learn more about what we do here in Studio, from game jams to our Student Run Studios program to fun socials. If you have any interest in developing games from start to finish, we're the committee for you! Whether it's programming, art, narrative, sound design, UI/UX, or any other aspect, you are welcome to join, regardless of skill level 🎵 🎨 💻 !! Come check us out, and let's make some games together! See you there 💖","links":[{"text":"Facebook Event","href":"https://www.instagram.com/p/CivIm1wJfyr/","ext":true}]},{"title":"Ludum Dare 51 Kickoff","allDay":false,"start":1664587800000,"end":1664595000000,"location":"Engineering VI 289","committee":"studio","description":"Ready to open the school year with a bang🤯 ? The biannual game jam, Ludum Dare is having its 51st jam this Friday, September 30th. The 🕐72 hour countdown🕐  begins once the jam's theme is revealed at 3:00 pm. Studio will be hosting a kickoff in E6 289 at 6:30 pm on Friday, so come join us and meet your team 💖 and get on the game jam grind! Don't worry if you missed the initial sign ups! You can sign up for the waitlist if you're interested in looking for a team through Studio: https://forms.gle/rNESbkHmRUXE3sMeA. Get ready to brainstorm, create, and realize a full game in just 72 hours!","links":[{"text":"Facebook Event","href":"https://www.instagram.com/p/Cissl1WPwkF/","ext":true}]},{"title":"Cyber Fall GM","allDay":false,
...
```
Output should either include the event or print an error for the event

# Notes
- Originally wrote the script in typescript, but I was having trouble compiling it in this repo, so I switched to javascript
[...]

PR #586, opened 2023-01-20
Fix error on events calendar week view

Amyh11325:
# Bug
Clicking on the week view in the events calendar throws an error.
Screen Shot 2023-01-20 at 12 46 05 PM

# Fix
Revert two lines from #563.
I initially thought the `new Date()` instantiation wasn't necessary. Turns out it's only used in non-month calendar views oops.

# Test
1) `yarn start`
2) go to `localhost:3000/events` 
3) Click on week view, month view, and agenda; Check that events appear as expected.
Screen Shot 2023-01-20 at 1 01 25 PM

### DIFF
--- a/pages/events.js
+++ b/pages/events.js
@@ -62,8 +62,8 @@ function Events({ events }) {
 						<Calendar
 							localizer={localizer}
 							events={indexedEvents}
-							startAccessor="start"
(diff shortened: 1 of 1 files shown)


matthewcn56 (approved):
Great, thanks for catching this!

PR #587, opened 2023-01-20
timezone issues resolved

matthewcn56:
addressed issues with timezones and moment.js


Amyh11325 (changes requested):
Nevermind still doesn't work on the deployed version rip

Amyh11325:
Fixed by modifying the timezone on netlify

Issue #478, opened 2022-04-11
Set up linters with Pre-Commit Hooks


matthewcn56:
As is, we can commit and make PRs before checking if our pages properly fit to our codebase's formatting standards. If we can make a github action to run `yarn lint-fix` and check to make sure that no errors occur on commit, we can prevent scenarios where we open PRs that won't build on netlify due to CI errors.

This can be achieved through husky precommit hooks that's similar to what's used in teach-la's [typescript react starter](https://github.com/uclaacm/teach-la-ts-react-starter)!

PR #593, opened 2023-01-31
Husky precommit hooks

ArshMalik02:
Implemented husky pre-commit hook for the website that checks linting and build. Some JS files were changed by eslint --fix during commit.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npm test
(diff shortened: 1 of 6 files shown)


smalex-z:
@ArshMalik02 is this related to #478 ? Should I look into merging this or should I just close it.?

ArshMalik02:
> @ArshMalik02 is this related to #478 ? Should I look into merging this or should I just close it.?

It indeed is related to #478. A bit of context: our linters were failing the builds because next image wasn't working back then, but with the next updates, we should have the linters working (hopefully) so the pre-commit hook can be enforced.

PR #603, opened 2023-03-07
Fix Github Action Failure from Missing Secrets

Amyh11325:
# Bug
Build and Test Github Action (running `npm build`) failing because Github secrets aren't being passed to the action.

# Fix
Add 
``` 
env:
        EVENTS_SPREADSHEET_ID: ${{ secrets.EVENTS_SPREADSHEET_ID }}
        SERVICE_ACCOUNT: ${{ secrets.SERVICE_ACCOUNT }}
```
to the .yml file to pass the secrets into the workflow

# Test
Github action passes when triggered from PR (see checks below)


bliutech (changes requested):
Might want to put the `SERVICE_ACCOUNT` and `EVENTS_SPREADSHEET_ID` into a GitHub Actions Secret. 👀  We don't want any unwanted haxors messing with things. 🙈 

More details below.
*Resource:* https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md**

Amyh11325:
> Might want to put the `SERVICE_ACCOUNT` and `EVENTS_SPREADSHEET_ID` into a GitHub Actions Secret. 👀 We don't want any unwanted haxors messing with things. 🙈
> 
> More details below. _Resource:_ https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md**

I'm just trying to get it to work T.T
I think these changes just invalidated the action which made the job not run

bliutech:
oop. 💀

ArshMalik02 (approved):

bliutech (changes requested):
I recommend rebasing your Git history. The API key is still in there and we don't want a bot to come scrape our key. 🙈 Probably a good idea to get a new API key as well.

ArshMalik02:
> I recommend rebasing your Git history. The API key is still in there and we don't want a bot to come scrape our key. 🙈 Probably a good idea to get a new API key as well.

if we squash and merge, then i believe it should handle that for us and it won't be in the git history? - matt

Amyh11325:
> I recommend rebasing your Git history. The API key is still in there and we don't want a bot to come scrape our key. 🙈 Probably a good idea to get a new API key as well.

Also we're going to delete that key so should be ok :v

bliutech:
> > I recommend rebasing your Git history. The API key is still in there and we don't want a bot to come scrape our key. 🙈 Probably a good idea to get a new API key as well.
> 
> Also we're going to delete that key so should be ok :v

Nice. This sounds like a good strategy to me. 👍 

> > I recommend rebasing your Git history. The API key is still in there and we don't want a bot to come scrape our key. 🙈 Probably a good idea to get a new API key as well.
> 
> if we squash and merge, then i believe it should handle that for us and it won't be in the git history? - matt

Recommend being cautious since even when merges are squashed for commits on the main branch, pull requests like this can still maintain commits. See this [example](https://github.com/uclaacm/website/pull/603/commits/36891ea8b5ecb586249a4b90817554a2d72d3bf4).

Issue #600, opened 2023-03-01
CI Problem Source


bliutech:
It appears there are some problems with default values on the builds for `Next.js` causing CI/CD to fail with `npm run build`. Highlighted two starting points below. Seems quite tricky. Hope this helps. 👍 

**Bug:**
https://github.com/uclaacm/website/blob/845076b4a99fad3d6b052f5815fa32a54449e239/scripts/event-generator-sheets.mjs#L9

**Potential Fix:**
```javascript
 const SPREADSHEET_ID = process.env.EVENTS_SPREADSHEET_ID ?? '';
```
`SPREADSHEET_ID` will be passed as undefined rather than an empty string type so the Google Sheets API can handle this.

**Bug:**
https://github.com/uclaacm/website/blob/845076b4a99fad3d6b052f5815fa32a54449e239/scripts/event-generator-sheets.mjs#L10

**Potential Fix:**
```javascript
 const SERVICE_ACCOUNT = process.env.SERVICE_ACCOUNT ?? '{}';
```

`JSON.parse` needs a sample JSON to fill here. I recommend finding an example for `private_key` and `client_email` attributes on this JSON so that there are default values. Private key is a little hard to generate so maybe the later usage should be ignored on static builds or something similar.

PR #610, opened 2023-04-16
Update banner on landing page

ArshMalik02:
ContentBanner.js: Edit component to display only text
index.js: add banner text
package.json: add postcss dep


jainsujay02 (changes requested):
The build preview failed. Is it failing because the tool we are using to test the deployment isn't configured correctly? Or is it because there's a problem with the code?

ArshMalik02:
@jainsujay02 the build is failing because of #600 bug that was reported earlier.

jainsujay02 (approved):
Approved with CI/CD not building preview.

PR #633, opened 2023-08-30
update Cyber links for new DNS

bliutech:
Cyber currently has a few new domain changes. Updating committee info to reflect this.

### DIFF
--- a/data/committees.js
+++ b/data/committees.js
@@ -226,7 +226,7 @@ export default [
 			},
 			{
 				platform: 'website',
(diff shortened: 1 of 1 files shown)


burturt (approved):

bliutech:
CI seems to be failing due to a misconfiguration within secrets rather than issues with this PR. @ArshMalik02 pinging for coverage. This can't be merged unless that is fixed and I don't believe I have access to wherever we store secrets.

PR #635, opened 2023-09-01
Added goat-counter analytics

jainsujay02:
- Added goat counter analytics. 
- Analytics can be viewed by an admin account owned by acm@ucla.edu


ArshMalik02 (approved):
lgtm

PR #679, opened 2024-01-25
Update landing page

ArshMalik02:
- index.js: Remove ContentBanner
- Footer.js: add Linktree button per request from marketing director


jainsujay02 (changes requested):
Instead of removing the banner, can we replace it with Fill out the ACM Census: https://forms.gle/hCTEKJvM7LnVKCez9

jainsujay02 (approved):

PR #723, opened 2024-10-16
Internship Info Update

smalex-z:
## Overview:

Updated internship info- members may apply to 3 committees instead of 2 now


snigdha-kansal (approved):
GIF

PR #736, opened 2024-11-19
Fix About Page Responsiveness

yeadamk:
## Overview

Changed media query min-width from 600px to 1160px, which is the navbar-breakpoint in `variables.scss`

## Changes

`styles/pages/About.module.scss`
- Change media query min-width to match navbar-breakpoint


snigdha-kansal (approved):
Looks great to me, thanks for doing this!

PR #783, opened 2025-04-14
impact site update

RohanAdwankar:
issue#780
it has some changes we wanted involving adding our ig feed, a carousel, etc
it would be great if we could get some help for merging this in!
thanks!


smalex-z:
Hi Rohan!

Thanks for the PR- We're currently in the midst of recruitment so I won't have time to fully analyze this PR until later next week, but a few things to tackle beforehand-

The deploy check is currently failing due to a bad google drive link that still matches our regex expression somewhere in the ACM Directory. This should be fixed in PR #784 , Feel free to pull that branch into yours and see whether that fixes the deploy checks failing.

Also, please take a minute to review and fix the ESLint errors. It tends to be rather picky about whitespace and trailing spaces, which tend to be the bulk of the warnings. There were a command to test ESLint compliance and one to try and automatically fix it, but I can't seem to remember it off the top of my head 😅. I'll send you a message if I remember the exact command.

smalex-z:
PR Moved to #791

PR #897, opened 2026-04-23
feat: add 2026-2027 officer year to alumni dropdown

smalex-z:
Add 2026-2027 to the officers page alumni years dropdown.

This allows viewing officers and staff from the 2026-2027 school year.


smalex-z:
Closing - created file from scratch instead of modifying existing file. Need to verify where alumyears.json actually comes from.

